PR #3, opened 2022-11-04
User/yreda/navigation bringup

younesr1:
this PR adds a simulated hospital world, a navigation pkg, and a manipulation pkg


dev_notes.txt
```diff
@@ -0,0 +1,31 @@
...
+launch gmapping with <roslaunch nightingale_navigation mapping.launch>
+drive the robot around and generate the map
+the map is publsihed to /map. you can visualize it through rviz
+save the map with <rosrun map_server map_saver -f 'map'>
+
+# Localization
+launch the map server and amcl with <roslaunch nightingale_navigation localization.launch>
+the /particlecloud topic shows the estimated pose array. you can visualize it through rviz
+
+younes todo add steps for the move base launch file
```
younesr1: ack of the todo

nightingale_navigation/configs/amcl.yaml
```diff
@@ -0,0 +1,52 @@
...
+gui_publish_rate:     -1.0      #default -1.0
+save_pose_rate:       0.5       #default 0.5
+use_map_topic:        false     #default false
+first_map_only:       false     #default false
+
+# Laser Model Parameters 
+laser_min_range:            -1.0    #default -1.0
+laser_max_range:            -1.0    #default -1.0
+laser_max_beams:            60     #default 30
+laser_z_hit:                0.90   #default 0.95
```
FelixKHWong, line 39: Are you able to explain the reasoning of why some parameters are changed from their defaults?

nightingale_navigation/configs/gmapping.yaml
```diff
@@ -0,0 +1,100 @@
...
+lasamplestep:              0.005   #default 0.005
+
+# How long (in seconds) between transform publications
+transform_publish_period:  0.05    #default 0.05
+
+#Threshold on gmapping's occupancy values. Cells with greater occupancy are considered occupied     
+occ_thresh:                0.25    #default 0.25
+
+#Initial map size
+xmin:                     -10.0   #default -100.0
```
FelixKHWong, line 89: These parameters seem to be very different than the defaults. Why is this?

FelixKHWong (dismissed):

FelixKHWong:
I mainly see a bunch of config files and they seem to be reasonable. I noticed that they all have their own launch file. Do these launch with Movo1 launch?

nightingale_navigation/configs/amcl.yaml
younesr1: so i took these params from https://github.com/nightingale-project/kinova-movo/blob/master/movo_demos/config/2d_nav/amcl.yaml

nightingale_navigation/configs/gmapping.yaml
younesr1: took em from. https://github.com/nightingale-project/kinova-movo/blob/master/movo_demos/config/mapping/gmapping.yaml im planning on spending more time understanding the params at some point in the future

younesr1:
> I mainly see a bunch of config files and they seem to be reasonable. I noticed that they all have their own launch file. Do these launch with Movo1 launch?

so each launch is for a specific use case. ill explain more in the readme. these nodes are purely information processing so they can run on any machine. they get all their data from ros topics and pub to ros topics (no IO)

younesr1:
younes todo move base rviz requires bas frame be map and that the goal pose publish to move_base_navi special topic not the default rviz has

nightingale_simulation/src/basic_arm_articulation.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+'''
```
younesr1: will delete this before merging

younesr1:
Here's a video showing MoveIt usage on the movo in simulation

https://user-images.githubusercontent.com/52610118/208826921-003b0b03-64c4-415b-ba60-e17a69df410b.mp4

nightingale_navigation/configs/amcl.yaml
younesr1: ive updated the params and now theyre different from what kinova uses. here's a summary:
max_particles increased from 2000 to 5000. I think we have extra compute to spare
laser min and max ranges now uses the actual values i found for our SIC laser
The rest of the laser model params are what kinova uses. See page 171 of Probabilistic Robotics for an explanation of the params. We use the likelihood_field_model
I also updated the motion model to use 'omni-corrected' instead of 'omni' since the latter had a [bug](https://github.com/ros-planning/navigation/issues/20). i also had to update those alpha constants as per [this post](https://answers.ros.org/question/227811/tuning-amcls-diff-corrected-and-omni-corrected-odom-models/)

nightingale_navigation/configs/gmapping.yaml
younesr1: decided not to dwell too much on understanding gmapping params since its not as important as params for amcl or move base. might come back to this, but the maps we generate seem pretty good.

nightingale_msgs/action/RoomRunner.action
```diff
@@ -0,0 +1,14 @@
+# Input is the patient name
+string patient_name
```
FelixKHWong: Would it be easier to have an alphanumeric ID (string) instead? Names can get complicated and be annoying to input if they are long. What about just using the initials of the patient and a number eg. Felix Wong -> FW2? We can also put in random characters too based on how hospitals do it

nightingale_navigation/README.md
```diff
@@ -0,0 +1,19 @@
...
+save the map with <rosrun map_server map_saver -f 'map'>
+
+# Localization
+launch the map server and amcl with <roslaunch nightingale_navigation localization.launch>
+the /particlecloud topic shows the estimated pose array. you can visualize it through rviz
+
+# Navigation Version 0
+launch nav with <roslaunch nightingale_navigation navigation.launch>
+if using sim, specify the arg location:=sim
+This launch file launces amcl, move base, and the room runner nodes
```
FelixKHWong, line 14: typo in launches

nightingale_navigation/maps/sim_hospital/dec.23.22/map.yaml
```diff
@@ -0,0 +1,7 @@
+image: map.pgm
+resolution: 0.050000
+origin: [-22.800000, -37.200000, 0.000000]
```
FelixKHWong, line 3: How important is the location of the origin? Is it just wherever the robot first appears in the map?

nightingale_navigation/maps/sim_hospital/nov11.22/map.yaml
```diff
@@ -0,0 +1,7 @@
+image: map.pgm
```
FelixKHWong, line 1: Should we remove these older maps since you have a december version?

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,288 @@
...
+            "Starting the Nightingale Navigation Room Runner Node!",
+            logger_name=self.logger_name,
+        )
+
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+
+        self.save_pose = rospy.get_param("/room_runner/save_pose", False)
+        self.pose_filename = rospy.get_param("/room_runner/pose_filename", "")
+        self.pose_write_period = rospy.get_param("/room_runner/pose_write_period", 60)
```
FelixKHWong: Why is the pose being written every 60 seconds? Does it provide a better estimate of where the robot it? Since we expect the robot to reach the room in 2.5 minutes should we be writing the pose more often maybe every 10 seconds? Or Is this milliseconds?

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,288 @@
+#!/usr/bin/env python3
```
FelixKHWong, line 1: Why is this file in python? Is it for easy implementation and to be ported to C++ later? Or will it stay as python?

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,288 @@
...
+
+        pub = rospy.Publisher("/initialpose", PoseWithCovarianceStamped, queue_size=10)
+        # Fast rate for checking if amcl is up
+        rate = rospy.Rate(100)
+        while pub.get_num_connections() == 0:
+            rospy.loginfo_throttle(5, "Waiting for amcl to subscribe to /initialpose")
+            rate.sleep()
+        pub.publish(msg)
+
+    def call_door_lookup(self, name):
```
FelixKHWong: What will this function do? Will it find all the known positions of doors? Or is it a function to engage the movement through a door

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,288 @@
...
+        ret.pose.position.x = 4.80
+        ret.pose.position.y = -3.34
+        ret.pose.position.z = 0.0
+        ret.pose.orientation.x = 0.0
+        ret.pose.orientation.y = 0.0
+        ret.pose.orientation.z = 0.0
+        ret.pose.orientation.w = 1.0
+        return ret
+
+    def read_pose(self):
```
FelixKHWong, line 187: Would it be useful to have some parameter to change the pose filename name? I'm wondering if it might be annoying to have to go through a file which contains all of the poses from every run and splitting up the files for the runs might make things more organized

FelixKHWong: After reading the code a bit more it seems like the pose is always overwritten so that there is only a single line in the file at a time?

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,288 @@
...
+        self.client.send_goal(msg, feedback_cb=fb_cb, active_cb=a_cb)
+        self.client.wait_for_result()
+
+        # move base result is empty so no need to check
+        result = RoomRunnerResult()
+        if self.client.get_state() == GoalStatus.SUCCEEDED:
+            rospy.loginfo(
+                "Nav to Room Completed Successfully :)", logger_name=self.logger_name
+            )
+        elif self.client.get_state() == GoalStatus.PREEMPTED:
```
FelixKHWong: When a call is preempted what happens? Right now does the robot just stop in place and wait?? Would it be useful to have the robot give up and go home or is it safest to stop and wait for a technician/someone to come fix the robot?

FelixKHWong:
Looks good! Good work!

nightingale_msgs/action/RoomRunner.action
younesr1: good point. ill change this to be room number which we assume to map uniquely to patient names, also the system which notifies us of a patient request likely knows the room number but not necessarily the patient name

nightingale_navigation/maps/sim_hospital/dec.23.22/map.yaml
younesr1: i think gmapping sets the map frame origin to the odom origin at the time of the mapping (or wherever u were when u launched gmappingg)

nightingale_navigation/maps/sim_hospital/nov11.22/map.yaml
younesr1: i think we can just keep em. i dont think theyre too large regenerating maps is pain, but ig we can get em off git history

nightingale_navigation/src/room_runner.py
younesr1: thats in seconds. though we write the pose on rospy.shutdown, we also write at some period just in case the node collapses suddenly or computer looses power or something unexpected. This is already pretty unlikely. The only reason we write the pose to disk is so that we have an initial guess when we wake back up. we could make it more frequent maybe 10 sec

nightingale_navigation/src/room_runner.py
younesr1: itll stay python. i tried cpp but was having some annoying linker error with on of actionlib's shared libs. the node doesnt do much intensive compute so python is fine. its just shuffling data

nightingale_navigation/src/room_runner.py
younesr1: this function should return the pose the robot should achieve prior to entering the room. In this pose, the robot should be facing the patient's door. this is hard-coded for now but will later call a service once lyndon makes the service

nightingale_navigation/src/room_runner.py
younesr1: second comment is correct. also the filename is a param specified in the launch file

nightingale_navigation/src/room_runner.py
younesr1: When an action is preempted, the robot cancels the move base action and stops moving. It is up to the caller to decide what to do next. If the caller would like the robot to go home, they should call the actino again with 'home' as the goal

nightingale_msgs/action/RoomRunner.action
FelixKHWong: I think patient id could work well rather than just the room number (maybe patient initials + room number to create the id?)  since we can use it to map to a string for "personalization" in the tablet interface.

nightingale_navigation/src/room_runner.py
FelixKHWong: If the pose is written on shutdown that's good

nightingale_navigation/src/room_runner.py
FelixKHWong: I see, what's  wrong with the pose being hard coded? If we are able to get the same start pose for entering the doorways wouldn't that make things easier to traverse through?

FelixKHWong (dismissed):

nightingale_navigation/src/room_runner.py
FelixKHWong: oh I misunderstood what you meant by hardcoded here

nightingale_msgs/action/RoomRunner.action
younesr1: for navigation purposes, i dont see why we need to know the patients name. we navigate the same way independent of the patient's name. for UI stuff, I agree it makes sense that we might want the patients name but this should be done in whatever interface node we have. that node should query name.

nightingale_msgs/action/RoomRunner.action
```diff
@@ -0,0 +1,16 @@
+# Input is the room number as string
```
younesr1, line 1: @lyndon just wanna let u see this action api

nightingale_navigation/src/room_runner.py
```diff
@@ -0,0 +1,234 @@
...
+        pub.publish(msg)
+
+    def lookup_pose(self, name, subloc):
+        # TODO Service Call, currently, just always returns the same pose
+        ret = PoseStamped()
+        ret.header.stamp = rospy.Time.now()
+        ret.header.frame_id = "map"
+        ret.pose.position.x = 4.80
+        ret.pose.position.z = 0.0
+        if subloc == "doorside":
```
FelixKHWong, line 177: What is the purpose of this if statement? is it just for testing and will be replaced with the service? What does the position under else represent?

FelixKHWong (approved):
Seems like most of the changes between last PR is just adding sim stuff which seems to work fine

nightingale_navigation/src/room_runner.py
younesr1: yes it is just for testing and will be replaced. i hardcoded arbitrary positions in the simulation map. in reality, this will call lyndons service and get a pose that way

PR #9, opened 2022-12-20
Felix kh wong add clang format

FelixKHWong:
Runs clang format check on pushes, and pull requests
Everyone will need to run "sudo apt install clang-format" to use clang format
https://github.com/marketplace/actions/clang-format-check


FelixKHWong:
Added github action for Clang format checking. As seen, the already formatted file passes and is not flagged while the unformatted file fails the check. Once approved I will go back and remove the test files then merge in the actual action file.

FelixKHWong:
Results of test with failure here: https://github.com/nightingale-project/nightingale/actions/runs/3742367589
Results of test without issues here: https://github.com/nightingale-project/nightingale/actions/runs/3762325912

younesr1 (approved):

younesr1:
dont forget to squash ur commits

PR #13, opened 2022-12-31
Felix/pt cloud lsr filter

### LINKED ISSUE #7: PointCloud -> LaserScan
Navigation should leverage camera point cloud in order not to drive through desks (or any surface not directly extruded from the groudn)
Navigation is planar and requires laser scans.

The idea is to use http://wiki.ros.org/pointcloud_to_laserscan to convert the point cloud to a laser scan and to fuse this laser scan with the other 2 laser scans.

The result of this ticket is a launch file placed under nightingale_navigation. The params for the node should make sense (ie. max height is max height of movo, angle increment is same as real lasers etc)

The author of this ticket should validate the correctness of the laser scan and that the freq is >= freq of real lasers. The correctness can be validated through gazebo and rviz by placing a table in front of the movo and ensuring it appears in the scan.

Maybe consider http://wiki.ros.org/depthimage_to_laserscan. though if throughput is a concern consider the nodelet api

FelixKHWong:
Integrates both the point cloud to laser scan package and the robot body filter package.

Point cloud to laser
- Converts part of the point cloud generated by the camera to a laser scan and projects it to the front laser link frame. This frame was chosen because it will not move relative to the robot as the robot moves unlike the camera frame which can telescope.
- Currently publishes to /movo/camera_scan. When publishing to the same topics as the laser scans, the scans swap between the camera projected scan and the base scans which would mess with navigation. Will look into laser scan merger packages so we can consider both datasets at the same time.
- Point cloud publishing slows down the sim. I get 8-10fps and I am concerned about the ability to run the sim with this package on other computers such as Lyndon's which is a bit older.

Below are tests with a chair and some computer stand in which the outer most part of the object is correctly projected as a laser scan,

converter_test
converter_test_2

Body filter
- Filters out the arms since only the arms are expected to come into camera line of sight.

Below are images of the body filter filtering out the arm when it is in view when the correct point cloud topic is used.

body_filtered
body_not_filtered


nightingale_visualization/package.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0"?>
```
FelixKHWong: Not sure why these changes are here. Might have to move my changes to a different branch to avoid this. Will take a look to fix

younesr1: kk just lmk when ur ready for a review

younesr1: i prob shouldve merged in stuff from my branch a while back. it pretty much has sim+arm+nav+misc now which is a mess. i might try to split things up soon

nightingale_perception/launch/pt_cloud_process.launch
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
```
FelixKHWong, line 1: rename to point_cloud_filtering.launch?

nightingale_visualization/package.xml
FelixKHWong: I think it's ready for review thanks! I remember you mentioned to go onto the nav-bringup branch to run sim which is why I branched off it so I can test the point cloud stuff in sim. Do I need to be on the nav-bringup branch to run sim or will any branch work? If I don't need to be off sim I'll move my stuff to avoid the conflicts to make things easier. I was able to update my local nav-bringup branch to get rid of most of the file differences but I still see one in the nav-sim folder. Is the nav-bringup branch branched off of a different branch? I'm trying to figure out where things are atm and where I should by my stuff

FelixKHWong:
Doing some more testing, using an object which cannot be seen or extrapolated, the point cloud converter works seen below.
However, I think there's some weird stuff going on with the scan merging and the scan filtering which I am looking into to see why the merge is not good and why in the filtered topic the camera scan vanishes. This might be due to the difference in angle increment which is causing the points to be seen through the gaps. I can change it and test

Also another issue I am looking into is that when I move the robot arm in the view of the camera when the body filter is on, the point cloud converter crashes and brings down the combined scan because a transform between the camera link and the front_last_link cannot be found which is strange

not_perfect_merge
scan_merger
scans_merged_base_filtered

FelixKHWong:
Using laser scan parameters for point cloud converter the behaviour is as expected and ready to go. However, there is an issue with the body filter with some lookup into the past. Otherwise very close to finished.

laser_scan_converter_working
scan_merge_same_params

FelixKHWong:
Doing some more debugging, it seems like the body filter might just be slow or maybe because I also have to run sim on my computer the performance is poor. The body filtered scans do appear but if I uncheck the Rviz pointcloud2 box, the cloud disappears. Rechecking and waiting, the filtered cloud eventually appeared but the same behaviour persisted

The blue dots projected into the laser scan are due to the mismatch with the urdf and the action robot position in which the gripper is actually open but in Rviz it isn't open. This behaviour is likely related to an issue with the grippers and not caused by this PR.

body_filter_Slow
filtered
unfiltered

FelixKHWong:
A possible solution to get this running faster would be to try to not consider the arm link closest to the shoulder since it doesn't seem like we'll raise it high enough to be seen.


arm_out

nightingale_perception/launch/pt_cloud_process.launch
```diff
@@ -0,0 +1,30 @@
...
+<launch>
+  <!-- launch file for point cloud laser scan conversion and filtering of robot body -->
+  <arg name="filtered_points_topic" default="/camera/depth/color/points_body_filtered"/>
+
+  <!-- Point cloud body filter node -->
+  <node name="pointcloud_filter" pkg="sensor_filters" type="pointcloud2_filter_chain">
```
younesr1: just fyi not specifying the screen attribute defaults to "log" for node xml elements. this means that logs from your nodes will not be printed to the terminal (only errors will be printed to terminal) just fyi in case u ex[ect useful logs

nightingale_perception/package.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>nightingale_perception</name>
+  <version>0.0.0</version>
+  <description>The nightingale_perception package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="felix@todo.todo">felix</maintainer>
```
younesr1: fyi

nightingale_perception/config/movo_body_filter.yaml
```diff
@@ -0,0 +1,25 @@
+cloud_filter_chain:
+- name: body_filter
+  type: robot_body_filter/RobotBodyFilterPointCloud2
+  params:
+     #frames/fixed: "odom"
+     frames/sensor: "camera_link" # might be camera_depth_optical_link
```
younesr1: does the data published by the camera node list the frame as camera_link or is it camera_depth_optical_link? i think this should match whichever it is?

nightingale_perception/config/movo_body_filter.yaml
```diff
@@ -0,0 +1,25 @@
...
+     sensor/point_by_point: False
+     sensor/min_distance: 0.05
+     body_model/inflation/scale: 1.07
+     body_model/inflation/padding: 0.03
```
younesr1, line 9: seems reasonable to me. pretty close to defaults. any reason were not using defautls? did u just use ur intuitiion?

nightingale_perception/config/movo_body_filter.yaml
```diff
@@ -0,0 +1,25 @@
...
+     filter/do_shadow/test: True
```
younesr1: filter/do_shadow_test (bool, default true)
If true, the filter will mark points shadowed by robot body as SHADOW. If false, such points will be marked OUTSIDE.

i dont think i quite understand the shadow test. why would we care about the shadow of the robot? especially since the output of this node is a laser scan which does not segment obstacles, right? is it just for visualization purposes?

It defaults to true so ig we want it?

nightingale_perception/config/movo_body_filter.yaml
```diff
@@ -0,0 +1,25 @@
+cloud_filter_chain:
+- name: body_filter
+  type: robot_body_filter/RobotBodyFilterPointCloud2
+  params:
+     #frames/fixed: "odom"
```
younesr1: nit: delte

nightingale_perception/config/movo_body_filter.yaml
FelixKHWong: I thought it would be good to be slightly conservative in case the urdf is not exactly as the real robot

nightingale_perception/config/movo_body_filter.yaml
FelixKHWong: From the tutorial in the constructsim it says that the shadow test is used to deal with noise caused around the arm by light refraction so I thought any points which appear behind the robot but should not be there would be removed. I can turn it off since it apparently is computationally heavy and if it becomes an issue I can turn it back on

Here' s the blurb if you want to read it cause otherwise you'd have to make the constructsim account and stuff:

Task 3: Shadow Filtering
Shadows are basically invalid points in free space that are generated by light refraction on edges of the robot's body. In simulation, there is nothing like refraction, but several settings can lead to these points being generated, too (like additive gaussian noise). Shadow filter is basically a raytracer that casts a ray from each measured point towards the sensor. If this ray goes through the robot's (inflated) body, the ray is considered invalid and is filtered out. Using the shadow filter allows to keep the collision model tight while still removing the shadow points relatively far from the robot body.

Raytracing is a relatively expensive computational operation, so always verify if your filter is realtime-capable if you enable shadow filtering. What helps is ignoring links that physically can't be seen by the lidar and thus cannot generate any shadow points. It also helps to keep the robot model simple and made of basic geometric shapes instead of meshes, or at least decreasing the mesh resolution. Also take into account that robot_body_filter always uses convex hulls of meshes, so if you use non-convex meshes, you have to either subdivide them into several convex parts, or count with the fact that it is the convex hull that is used. There is also a setting for maximum distance of the points considered for shadow filtering, and this can help a lot if a large portion of the pointclouds is far from the sensor (we assume the body-generated shadow points are all located nearby the robot's body).

nightingale_perception/config/movo_body_filter.yaml
FelixKHWong: Got catch the frame in the pointcloud data is "camera_depth_optical_frame". I'll be more thorough by making a checklist and keeping better tabs on things I need to do before PR sorry.

nightingale_perception/config/movo_body_filter.yaml
FelixKHWong: Tested with shadow test False and the performance seems around the same so I'll leave it off

nightingale_perception/launch/pt_cloud_process.launch
```diff
@@ -0,0 +1,30 @@
...
+  </node>
+
+  <!-- Point cloud to laser scan conversion -->
+  <node pkg="pointcloud_to_laserscan" type="pointcloud_to_laserscan_node" name="pointcloud_to_laserscan_node" output="screen">
+      <remap from="/cloud_in" to="$(arg filtered_points_topic)"/>
+      <!--remap from="/cloud_in" to="/camera/depth/color/points"/-->
+      <remap from="/scan" to="/movo/camera_scan"/>
+
+      <rosparam>
+          target_frame: "front_laser_link" #front laser link chosen for
```
FelixKHWong, line 21: After reviewing the ROS bag data taken before, the laser parameters in the real front laser are slightly off. May need to add an if statement here to check for "MOVO_SIM=true" to use different params. Will test in person
real_front_laser_params

FelixKHWong:
Did some testing in the lab and the point cloud to laser scan works well shown below.

point_cloud_to_laser working

There is a bit of noise due to the realsense that is also converted to laser scans. The realsense Ir depth is not perfect and creates some wavy artifacts which I'm not sure if there is a way to filter them out. The noise does not appear to be too significant so I don't think it will heavily impact navigation.

noise_camera_scan

The main issue is the body filter which still has poor performance on the real robot. Shown below is the body filter in action. The arm is in front of the camera and the point clouds are being filtered out. There are still some points because moveit thinks the gripper is open when in reality it is actually closed. None the less the remainder of the arm is mostly filtered out and the filtering box can be expanded out to make sure we don't accidentally leave points. However, the rate is still very low at around 0.08Hz. Looking at the documentation I can try to turn off the shadow test(which I'm not sure if I pulled the most recent version of my code which should have it off) and remove the arm links closest to the body since I don't think its likely we will be manipulating the arms to be in from of the camera like that. I also talked to Dr DeHart about it and he suggested that the body filter might not be needed at all times. If we expect to tuck the arms to stow the bin, the arms will never impact the point cloud to laser scan. During bin handover, moveit already knows where the arms are so would it be necessary to filter them out of the point cloud? The point cloud to laser scan may not be necessary during bin handover since it is used mainly for navigating around obstacles which is not needed for bin handover beside the bed. If we need the point cloud for object tracking we could possibly use the image data to help identify objects and then combine them with the point cloud to accurately identify objects that are too close rather than filtering out the arms. 
The body filter is working but I'm not sure if I can get it to realtime performance with the cpu we have. I'll try it again tomorrow after before demos.


body_filtered_irl

non filtered. Can see the actual arm texture on the Rviz arms.
body_not_filtered_irl

FelixKHWong:
After more investigation I've found that there was a spelling mistake in the config file for the setting of the shadow test. With shadow test off the filter runs quickly and does not seem to impact the performance much.

nightingale_perception/launch/pt_cloud_process.launch
younesr1: thats fine

nightingale_perception/README.md
```diff
@@ -0,0 +1,9 @@
+# Perception
+launch point cloud body filtering and point cloud to laser scan conversion with <roslaunch nightingale_perception pt_cloud_process.launch>
+body filtered points topic is /camera/depth/color/points_body_filtered
+point cloud converted to laser scan topic is /movo/camera_scan
+
+# Map Frame Visualizer
```
younesr1: did u mean to push this? map frame viz is not in nightingale_perception

younesr1 (dismissed):
good to merge after fixing the commend i left

nightingale_perception/README.md
FelixKHWong: removed

nightingale_perception/launch/pt_cloud_process.launch
FelixKHWong: Testing in real world with current params seems fine performance

FelixKHWong:
@younesr1 can I get another approval here? I forgot that with the laser scan merger change to include the camera the combined scans will not publish unless this PR goes in. So currently master does not have a working laser scan since it requires the point cloud to laser scan node to be run. Alternatively I could push a quick change onto Kinova movo to remove the change to laser scan merger and then move this PR over to Kinova MOVO and have it launch as part of MOVO2

younesr1 (dismissed):
reapproving

FelixKHWong:
Added in merger node to launch file. Ready to be merged now

nightingale_perception/launch/pt_cloud_process.launch
```diff
@@ -0,0 +1,49 @@
...
+          max_height: 0.20 # determined using trial and error in sim to see which value gives points and not ground plane
+      </rosparam>
+  </node>
+
+  <!-- Laser scan merger for camera scan -->
+    <node pkg="ira_laser_tools" name="laserscan_multi_merger_camera" type="laserscan_multi_merger" output="screen">
+        <param name="destination_frame" value="base_link"/>
+        <param name="cloud_destination_topic" value="/movo/merged_cloud_camera"/>
+        <param name="scan_destination_topic" value="/movo/scan_multi_camera"/>
+        <param name="laserscan_topics" value ="/movo/$(optenv LASER1_PREFIX front)_scan /movo/$(optenv LASER2_PREFIX rear)_scan /movo/camera_scan"/>
```
younesr1: why remerge rear_scan and front_scan? why not just merge merged_scan and camera_scan?

nightingale_perception/launch/pt_cloud_process.launch
```diff
@@ -0,0 +1,49 @@
...
+    </node>
+
+    <node pkg="laser_filters" type="scan_to_scan_filter_chain" name="movo_laser_filter_camera" clear_params="true" respawn="true">
```
younesr1: why not output=screen?

nightingale_perception/config/movo_body_filter.yaml
```diff
@@ -0,0 +1,25 @@
...
+     sensor/min_distance: 0.05
+     body_model/inflation/scale: 1.07
+     body_model/inflation/padding: 0.03
+     filter/do_shadow_test: False
+     filter/max_shadow_distance: 2.0
+     transforms/buffer_length: 15.0
+     transforms/timeout/reachable: 0.2
+     transforms/timeout/unreachable: 0.2
+     #filter/model_pose_update_interval: 0.005 # arbitrary update interval selected for now
+     only_links: ["left_arm_half_2_link", "left_forearm_link", "left_gripper_base_link",
```
bdschnapp, line 16: Any reason to include shoulder links here? Will the camera ever see the shoulder? Otherwise I think this list of links is correct

FelixKHWong: The shoulder join has its own link which is not listed here. I wasn't sure if the long joint closest to the shoulder would ever cross the path of the camera but I left it in just to be safe.

FelixKHWong:
I went back and did some testing with merging the camera scan ontop of the /scan_multi and found some strange results that merging the front, rear and camera at once has better results than merging the camera ontop of the merged laser scans. shown here:
1. Merging all together
base_scan_together
The reason is the ScanShadow filtering as when I remove it the base_scan_filtered scan looks similar though I'm not quite sure why. I changed the number of neighbours to be less for the filter and got better results but that could be because the filter has less effect.

filtered_camera_series
filtered_camera_all

I read the source code of the filter but I still couldn't figure out the reason for the differences. The filter checks X number of neighbours in a scan and then removes the neighbours which are farther away though I'm not sure why that is. The points being removed by the filter are not "in the middle" as stated in the documentation http://wiki.ros.org/laser_filters/Tutorials/Laser%20filtering%20using%20the%20filter%20nodes. My best guess is a time delay between the merged scan and then merging another scan ontop of it. I don't think this is worth digging into more unless we have problems with the scan since it seems to work similar to before.

younesr1:
> I went back and did some testing with merging the camera scan ontop of the /scan_multi and found some strange results that merging the front, rear and camera at once has better results than merging the camera ontop of the merged laser scans. shown here:
> 
> 1. Merging all together
>    base_scan_together
>    The reason is the ScanShadow filtering as when I remove it the base_scan_filtered scan looks similar though I'm not quite sure why. I changed the number of neighbours to be less for the filter and got better results but that could be because the filter has less effect.
> 
> filtered_camera_series filtered_camera_all
> 
> I read the source code of the filter but I still couldn't figure out the reason for the differences. The filter checks X number of neighbours in a scan and then removes the neighbours which are farther away though I'm not sure why that is. The points being removed by the filter are not "in the middle" as stated in the documentation http://wiki.ros.org/laser_filters/Tutorials/Laser%20filtering%20using%20the%20filter%20nodes. My best guess is a time delay between the merged scan and then merging another scan ontop of it. I don't think this is worth digging into more unless we have problems with the scan since it seems to work similar to before.

nav will not use the filtered versions of the scans

younesr1 (approved):

PR #16, opened 2023-01-02
Felix/create interface files

FelixKHWong:
Hoping to merge at that point so that others can hop in and work ontop if they have changes they want to make on the interface

This PR contains the python code for the user tablet interface for the MOVO robot. It has multiple screens to host the interactions for the patient. Most screens do not have ROS interaction. Once a patient completes an action that requires ROS/robot interaction, they will be sent to a confirmation screen to confirm the action and is "yes", a message will be published to a topic. Most of the ROS topic publishing will be in the "confirmation_screen.py" file and the listening to the ROS master will be in the "main.py" file where the thread can execute actions regardless of what is on the screen.

Some screens (The screens with arm interaction) have not been fully fleshed out yet since they will require topics to be published for the  interface to listen to so the robot is in the right place to execute on them. The designs for these screens have been completed but the functionality has not been fully fleshed out yet.

Sample images are here:

admin
home
itemrequest

alert
confirmation
estop
face

PR #17, opened 2023-01-10
Ben/interface ros v0

bdschnapp:
created simple backend for the tablet interface using ros bridge. 

Known issue: launch file does not work, and rosbridgde server must be launched seperatly from ros_bridge_interface ros_bridge_interface.py


nightingale_interface/main.py
```diff
@@ -84,14 +86,28 @@ def init_ros(self):
...
-        self.dummy_topic = roslibpy.Topic(
-            self.client, "/qt_robot/behavior/talkText", "std_msgs/String"
-        )
-
-    def listen_ros_master(self):
-        # function to wait for responses from the master
-        pass
+        self.ros_action_topic = roslibpy.Topic(self.client, "INSERT/TOPIC/HERE", "std_msgs/String")
+
+        self.interface_screen_topic = roslibpy.Topic(self.client, "interface/set_screen", "std_msgs/String")
```
FelixKHWong: what does this mean? Is it the topic will control the screen changes?

nightingale_interface/screens/screen_config.py
```diff
@@ -9,8 +10,8 @@ class ScreenConfig:
     INCR_BUTTON_XPOS = 0.75
     INCR_BUTTON_FONTSIZE = 50
 
-    CURRENT_ACTION = "Waiting"
-    PENDING_ACTION = "Waiting"
+    CURRENT_ACTION = 0
```
FelixKHWong: What will these numbers represent? i was thinking that the variables "CURRENT_ACTION" and "PENDING_ACTION" would be the strings that we can check and also show on screen for more information. Do you think we should keep the integers and have a lookup table for strings if we need to show the action on screen?

ros_bridge_interface/package.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>ros_bridge_interface</name>
+  <version>0.0.0</version>
+  <description>The ros_bridge_interface package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="ben@todo.todo">ben</maintainer>
```
FelixKHWong: replace with your email

ros_bridge_interface/package.xml
```diff
@@ -0,0 +1,65 @@
...
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
FelixKHWong: MIT

ros_bridge_interface/package.xml
```diff
@@ -0,0 +1,65 @@
...
+  <!--   <test_depend>gtest</test_depend> -->
+  <!-- Use doc_depend for packages you need only for building documentation: -->
+  <!--   <doc_depend>doxygen</doc_depend> -->
+  <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>rospy</build_depend>
+  <build_depend>std_msgs</build_depend>
+  <build_export_depend>rospy</build_export_depend>
+  <build_export_depend>std_msgs</build_export_depend>
+  <exec_depend>rospy</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
```
FelixKHWong: do we need to add rosbridge here as a dependency to run?

nightingale_interface/main.py
```diff
@@ -102,8 +103,13 @@ def call_ros_action(self, action: int, args: dict = {}) -> bool:
             args = {}
         args['action'] = str(action)
         json_str = json.dumps(args)
+        print(str(json_str))
```
FelixKHWong: do we want this print statement?

ros_bridge_interface/launch/interface.launch
```diff
@@ -0,0 +1,6 @@
+<launch>
+  <include file="$(find rosbridge_server)/launch/rosbridge_websocket.launch" >
+     <arg name="port" value="9091"/>
```
FelixKHWong: are these arguments default? Or will we need to change the port later?

FelixKHWong (changes requested):
Can you look at the previous PR here https://github.com/nightingale-project/nightingale/pull/16 ? I think I won't approve this until the previous PR gets merged in

ros_bridge_interface/launch/interface.launch
bdschnapp: This is the port that is set in MovoConfig. I believe the default is 9090. I would like to make this value a parameter and controlled from one spot eventually but didn't have time yet.

nightingale_interface/main.py
bdschnapp: good catch, that was for debugging

ros_bridge_interface/package.xml
bdschnapp: nope, rosbridge server runs in it's own package and this package in just subscribing to the topic that rosbridge is publishing to

nightingale_interface/screens/screen_config.py
bdschnapp: My understanding of your code was that these were supposed to be enums, sorry if I missunderstood. It makes the most sense to use enums when sending data through

nightingale_interface/main.py
bdschnapp: yes this is the screen change topic

dev_notes.txt
```diff
@@ -2,9 +2,13 @@
 run <roslaunch nightingale_teleop keyboard.launch> to move robot with keyboard
 this automatically sets the robot in tractor mode
 
+# Tablet interface
```
FelixKHWong: Create a new README file for this. I can probably do this on the previous PR

nightingale_interface/main.py
```diff
@@ -29,6 +29,7 @@ class MainApp(MDApp, ScreenWrapper):
     # task queue things
     queue_task = []
```
FelixKHWong: Why are the task and the args(I'm assuming data for each data) split up? Wouldn't it be best if the task is a single object which has fields containing the task ID and also fields with data for that task?

nightingale_interface/main.py
```diff
@@ -81,15 +89,14 @@ def build(self):
...
         self.client = roslibpy.Ros(
-            "localhost", MovoConfig.Config["RosBridgePort"]
+            MovoConfig.Config["Movo2"]["IP"], MovoConfig.Config["RosBridgePort"]
         )
-        #MovoConfig.Config["Movo2"]["IP"]
         self.client.run()
         asyncio.sleep(0.5)
 
-        self.ros_action_topic = roslibpy.Topic(self.client, "interface/action", "std_msgs/String")
+        self.ros_action_topic = roslibpy.Topic(self.client, "ui/robot/call_action", "std_msgs/String")
```
FelixKHWong: what do these action and interface topics do? I was thinking we would subscribe to a single status topic and make decisions based on the message recieved

ros_bridge_interface/src/ros_bridge_interface.py
```diff
@@ -0,0 +1,35 @@
...
+import rospy
+from std_msgs.msg import String
+import json
+import time
+
+
+class RosBridgeInterface:
+    def __init__(self):
+        rospy.init_node('ros_bridge_interface_server', anonymous=True)
+        self.interface_action_sub = rospy.Subscriber("ui/robot/call_action", String, self.interface_action_callback)
```
FelixKHWong: Can we change the topic the interface node subscribes to to "nightingale_ui/user_input"? I think this makes more sense since it is listening to messages the tablet is sending for when the patient wants the robot to do something

ros_bridge_interface/src/ros_bridge_interface.py
```diff
@@ -0,0 +1,35 @@
...
+        self.set_screen_pub = rospy.Publisher("ui/app/set_screen", String, queue_size=10)
```
FelixKHWong: Can we change the interface node published topic to "nightingale_ui/robot_status"? I think this makes more sense since the interface is relaying changes in the robot status for the tablet to reach to by changing the screen.

ros_bridge_interface/src/ros_bridge_interface.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def interface_action_callback(self, msg):
```
FelixKHWong: I see this is the callback for the interface to react to the data sent from the tablet. Where does this send the data back to the mission planner? Will there be a service called on the interface node which kicks off the publisher to send the status to the tablet and then the tablet responses with user input and then input is then sent back as a response for the service?

nightingale_interface/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+        async def run_wrapper():
+            await self.async_run(async_lib="asyncio")
+            print("App done")
+            self._other_task.cancel()
+
+        return asyncio.gather(run_wrapper(), self._other_task)
+
+    # add a task to do after a delay
+    def queue(self, task, delay=0, args=None):
```
FelixKHWong, line 66: Now that we've discussed the system as sending a status to the UI and the UI sending the user input data back, is this "queue" functionality required? i feel like the mission planner will always send the next status when it is available and will not end up queuing multiple tasks to do at a time.

nightingale_interface/main.py
```diff
@@ -0,0 +1,147 @@
...
+        # initialize the ros bridge client
+        self.client = roslibpy.Ros(
+            MovoConfig.Config["Movo2"]["IP"], MovoConfig.Config["RosBridgePort"]
+        )
+        self.client.run()
+        asyncio.sleep(0.5)
+
+        self.ros_action_topic = roslibpy.Topic(self.client, "ui/robot/call_action", "std_msgs/String")
+
+        self.interface_screen_topic = roslibpy.Topic(self.client, "ui/app/set_screen", "std_msgs/String")
```
FelixKHWong: requested topic name change above

nightingale_interface/main.py
```diff
@@ -0,0 +1,147 @@
...
+        msg = roslibpy.Message({"data": json_str})
+        try:
+            self.ros_action_topic.publish(msg)
+            return True
+        except:
+            return False
+
+    def set_screen_callback(self, msg):
+        # if the screen requested exists
+        if self.get_screen(str(msg['data'])):
```
FelixKHWong: Should now be based on the "status" received

FelixKHWong (changes requested):
I can also go onto your branch and make the changes if that's ok with you

nightingale_interface/main.py
bdschnapp: thats a good point, this should be a single variable instead of three and just use a dict to pass everything in to it

nightingale_interface/main.py
bdschnapp: This is more for blocking tasks to keep the touch screen unblocked rather than to actually queue tasks. This allows the execution of blocking tasks from a seperate thread

ros_bridge_interface/src/ros_bridge_interface.py
bdschnapp: The links between rb interface and mission planner need to be implemented. Services, actions, and topics that communicate with mission planner should be defined here, and the data recieved in this callback should be processed and forwarded to the correct service/action/topic

nightingale_interface/screens/screen_config.py
FelixKHWong: We'll use enums and then convert to strings later if needed

nightingale_interface/main.py
```diff
@@ -44,24 +52,37 @@ async def run_wrapper():
...
-                    func = self.queue_task.pop()
-                    await asyncio.sleep(self.queue_delay.pop())
-                    func()
+                if len(self.task_queue):
+                    current = self.task_queue.pop() 
+                    func = current["task"]
+                    await asyncio.sleep(current["delay"])
+                    func_args = current["args"]
+
+                    if args:
```
bdschnapp: "if func_args:" ? I don't see args defined here and I think this will cause a run time error

nightingale_ros_bridge/launch/interface.launch
```diff
@@ -0,0 +1,6 @@
+<launch>
```
bdschnapp, line 1: Has this launch file been tested yet? Last time I used it it didn't seem to work and I had to launch the nodes individually

nightingale_ros_bridge/src/nightingale_ros_bridge/bridge_interface_config.py
```diff
@@ -0,0 +1,35 @@
...
+    NO_ROS_ACTION = -1
+    STOCK_ITEMS = 9
+    DELIVER_ITEMS = 10 
+    RETURN_HOME = 11
+    START_EXTEND_ARM = 12
+    ITEMS_TAKEN = 13
+    START_RETRACT_ARM = 14
+    ESTOP = 15
+    ESTOP_CANCEL = 16
+    WD_TIMEOUT = 17
```
bdschnapp: Let's talk about this in person, I understand why this was done this way, but I don't like it and am having a hard time explaining why I don't like it

bdschnapp (commented):
Looks good to me. Only the first comment needs to be fixed before merging. The other comments can be discussed later

nightingale_interface/main.py
FelixKHWong: I think we should remove this now that the Mission planner is doing the actual computation and the tablet is just an input interface. What do you think?

nightingale_ros_bridge/src/nightingale_ros_bridge/bridge_interface_config.py
FelixKHWong: Yeah is it because it doesn't start at 1? I was concerned there would be mistakes made from overlap with the Robotstates. Having these start at 0 also makes sense since they come from different class objects

nightingale_interface/main.py
FelixKHWong: oh yeah this is a typo good catch. Do you think we should just remove this queue task stuff? It seems like the mission planner will be doing the task queue stuff and that the tablet is just being used as an interface

nightingale_ros_bridge/launch/interface.launch
FelixKHWong: I think so, I'm running "roslaunch nightingale_ros_bridge interface.launch" and getting things communicating as you saw today

nightingale_interface/main.py
```diff
@@ -9,59 +11,94 @@
...
+        self.screen_stack.append(self.root.current)
 
         try:
             while True:
 
+                # update screen upon new state
+                # need better way. probably just another state
+                if len(self.ros_set_screen) > 6:
+                    self.root.current = self.ros_set_screen
+                    self.ros_set_screen = ""
```
bdschnapp: This is dangerous code, any value could get put into self.ros_set_screen and if it is not a screen name the app will crash. Use something like: 
`if self.ros_set_screen and self.get_screen(self.ros_set_screen):`
`    self.root.current = self.ros_set_screen`
`    self.ros_set_screen = None`

the get_screen method is from the screen_wrapper class

nightingale_interface/main.py
bdschnapp: I disagree. Lets talk in person about this, it has made me think of more things that need to be implemented specifically related to the estop

nightingale_interface/screens/home_screen.py
```diff
@@ -8,50 +8,70 @@
...
-class HomeScreen:
-    home_name = "homescreen"
+sys.path.append("..")
+# TEMPORARY UNTIL INCLUDES ARE FIXED
+
+from nightingale_ros_bridge.src.nightingale_ros_bridge.bridge_interface_config import (
+    BridgeConfig,
+    UserInputs,
+)
+
```
bdschnapp: This is a good temporary fix, eventaully we can package the app to run as an executable though and the imports will be handeled in the build process. https://kivy.org/doc/stable/guide/packaging-windows.html

bdschnapp (commented):
looks good! Only one comment needs to be fixed, I don't think this will need another review before merging

nightingale_interface/main.py
FelixKHWong: ok I'll leave it in for V0

FelixKHWong (dismissed):
Should be ready to complete V0

nightingale_interface/main.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import roslibpy
+import json
```
younesr1: consider running a static analyzer like pylint https://pylint.pycqa.org/en/latest/user_guide/usage/run.html

bdschnapp (commented):
Looks good!

nightingale_interface/main.py
FelixKHWong: I ran the linter and my code is pretty bad though with how Kivy works I think the score would be quite low regardless since it allows me to use variables in other classes which is normally not possible. It would take me too long to refactor and make things much better and I feel that the code is readable enough for now. I did use the linter to clean up some errors such as the import errors for now.

FelixKHWong (approved):

PR #20, opened 2023-01-16
User/le2tang/room lookup0

le2tang:
Add a service "/lookup_room_pose" to match features in the room to a pose in the map


younesr1:
can this be in a new pkg called nightingale_utils? can u also make a launch file for nightingale_utils which runs this node s.t the service is advertised

nightingale_msgs/srv/RoomPoseLookup.srv
```diff
@@ -0,0 +1,8 @@
+string room
+string feature
+---
+int8 SUCCESS=0
```
younesr1, line 4: is this better than putting the return codes as a comment and having one int instead of 4?

le2tang: If you are asking if we can return a string with a return status message, that implementation also works but seems a bit excessive. If you are asking why there are 4 int8 fields for the return status message, these are just constants as shown here: https://wiki.ros.org/msg#Constants. The use case I am using is the autogenerated enum in Python (ie. RoomPoseLookup.ROOM_NOT_FOUND).

nightingale_utils/src/room_pose_lookup.py
```diff
@@ -0,0 +1,66 @@
...
+            "room_pose_lookup", RoomPoseLookup, self.room_pose_lookup
+        )
+
+    def load_rooms(self):
+        rooms = {}
+
+        rooms_params = rospy.get_param("rooms")
+
+        for room in rooms_params.keys():
+            features = {}
```
FelixKHWong, line 25: Why do you use the term "features" here? Is this in preparation of some other custom traversal behaviour for each patient? I thought that we would only need the position and orientation of the robot goals since that is what we need to plan a path to the room.

nightingale_utils/launch/room_pose_lookup.launch
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<launch>
+    <node pkg="nightingale_utils" type="room_pose_lookup.py" name="room_pose_lookup">
```
FelixKHWong: could add "output=screen" for more visibility of whats going on

nightingale_navigation/maps/lab/jan.12.23/map.yaml
```diff
@@ -0,0 +1,7 @@
+image: map.pgm
```
younesr1: did u mean to push this?

nightingale_utils/src/room_pose_lookup.py
younesr1: each room has an associated feature. a feature is one of {door,bedside}. this allows us to first navigate to right infront of the door, (possibly do other work like opening a door), and then going through the door and to the bedside.

This makes adding code for opening doors easier to integrate

nightingale_utils/src/room_pose_lookup.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
```
younesr1, line 2: tip: consider running static analyzer on this script like pylint <pylint room_pose_lookup.py>. i find useful as it catches misspelled vars and just dumb mistakes generally

nightingale_msgs/srv/RoomPoseLookup.srv
younesr1: i meant second and sure nw

younesr1 (dismissed):
feel free to merge if tested

nightingale_navigation/maps/lab/jan.12.23/map.yaml
younesr1: - [ ] todo

nightingale_msgs/srv/RoomPoseLookup.srv
bdschnapp: Just confirming for my understanding lines 4-6 are just defining constants, then line 7 is the actual field that data will get sent through?

le2tang: Correct

younesr1 (approved):

PR #22, opened 2023-01-18
tuning navigation for e5 lab

younesr1:


le2tang (dismissed):
If you could add a comment in the PR to explain the choice from moving from base_scan_filtered to scan_multi, that would be helpful for documenting changes. Everything else looks good.

nightingale_dispatcher/src/nightingale_dispatcher/dispatcher.py
```diff
@@ -36,8 +36,9 @@ def handle_add_mission(self, req):
             self.mission_planner_client.send_goal(
                 self.missions.get(), self.handle_done_task
             )
-
-        return True
+        ret = AddMissionPlanResponse()
+        ret.status = True
+        return ret
```
younesr1, line 41: fyi @le2tang made this bool for now bc i dont think error codes r assigned yet. feel free to change once they are

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -24,17 +24,18 @@ def __init__(self):
...
-        )
-        self.server.start()
 
         self.handoff_task = HandoffTask()
         self.idle_task = IdleTask()
         self.navigate_task = NavigateTask()
         self.stock_task = StockTask()
         self.triage_task = TriageTask()
 
+        self.server = actionlib.SimpleActionServer(
```
younesr1, line 29: @le2tang i think action server should be launched after the tasks are initialized bc the constructors block as they wait for the action. So what i saw happen was I ran dispatcher which created this action and then created the navigate_task instance. I'd forgotten to run room runner so the navigate_task constructor kept waiting for the room runner action. Then I ran the AddMission service and got a python error bc the navigate_task was not defined since the constructor was stil lwaiting

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -104,12 +105,14 @@ def go_idle(self):
         if status == Task.ERROR:
             raise NotImplementedError()
 
+    # younes todo ask lyndon if this isnt a race condition
     def goal_cb(self, goal):
```
younesr1: @le2tang here u assume a 2nd action is not called before the 1st action ends correct?

PR #23, opened 2023-01-19
User/le2tang/dispatcher0

le2tang:
Add basic outline of task manager
Task is one delivery for a single patient composed of a set order of subtasks
Subtask is one high level action like navigation and bedside interaction


nightingale_dispatcher/package.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>nightingale_dispatcher</name>
+  <version>0.0.0</version>
+  <description>The nightingale_dispatcher package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="lyndon@todo.todo">lyndon</maintainer>
```
FelixKHWong: put your email here

nightingale_dispatcher/package.xml
```diff
@@ -0,0 +1,65 @@
...
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
FelixKHWong: MIT?

nightingale_dispatcher/src/nightingale_dispatcher/main.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+
+import rospy
+
+import queue
+from nightingale_msgs.srv import AddTaskDispatcher, CancelTaskDispatcher
+from nightingale_dispatcher.subtask import Subtask
```
FelixKHWong: So tasks(functions which actually execute behaviour)  are called "subtasks"? Why not call them tasks?

nightingale_dispatcher/src/nightingale_dispatcher/main.py
```diff
@@ -0,0 +1,53 @@
...
+from nightingale_dispatcher.navigate_subtask import NavigateSubtask
+
+
+class DispatcherNode:
```
FelixKHWong: Are there ways to pre-empt/pause tasks?  I'm thinking because for the hand bin tasks we will need user input which will require waiting on the tablet input and calling of the "interface call" task. Unless the connection to the interface is not to be in a task?

nightingale_dispatcher/src/nightingale_dispatcher/navigate_subtask.py
```diff
@@ -0,0 +1,31 @@
...
+class NavigateSubtask(Subtask):
+    def __init__(self, name, room, feature, priority):
+        super.__init__(name, priority)
+
+        self.room = room
+        self.feature = feature
+
+        # TODO Consider making a Singleton or share resources
+        rospy.wait_for_service("/lookup_room_pose")
+        self.lookup_proxy = rospy.ServiceProxy("/lookup_room_pose", RoomLookup)
```
FelixKHWong: Is the type here supposed to be "RoomPoseLookup"? There doesn't seem to be any service with name RoomLookup

nightingale_dispatcher/src/nightingale_dispatcher/navigate_subtask.py
```diff
@@ -0,0 +1,31 @@
...
+
+class NavigateSubtask(Subtask):
+    def __init__(self, name, room, feature, priority):
+        super.__init__(name, priority)
+
+        self.room = room
+        self.feature = feature
+
+        # TODO Consider making a Singleton or share resources
+        rospy.wait_for_service("/lookup_room_pose")
```
FelixKHWong: Have the service name as a constant which can be called rather than manually putting it in?

nightingale_dispatcher/src/nightingale_dispatcher/main.py
```diff
@@ -0,0 +1,51 @@
...
+from actionlib.simple_action_client import SimpleGoalState
+from nightingale_msgs.msg import TaskAction, TaskGoal
+from nightingale_msgs.srv import AddTaskDispatcher
+
+
+class DispatcherNode:
+    def __init__(self):
+        rospy.init_node("dispatcher_node")
+
+        self.task_client = actionlib.SimpleActionClient("task", TaskAction)
```
younesr1: can u add logs aroudn the wait 4 server. can be hard to debug bc itll just sit there

nightingale_dispatcher/src/nightingale_dispatcher/subtask.py
```diff
@@ -0,0 +1,22 @@
...
+        self.priority = priority
+
+    def __lt__(self, other):
+        return self.priority < other.priority
+
+    def __eq__(self, other):
+        return self.priority == other.priority
+
+    def execute(self):
+        pass
```
younesr1:
```suggestion
    def execute(self):
        raise NotImplementedError()
```

nightingale_dispatcher/launch/dispatcher.launch
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0"?>
+<launch>
+    <node pkg="nightingale_dispatcher" type="main.py" name="dispatcher" output="screen"></node>
```
younesr1: do there rly need to be 2 processes? could they be combined

nightingale_dispatcher/src/nightingale_dispatcher/task_action_server.py
```diff
@@ -0,0 +1,34 @@
...
+from actionlib_msgs.msg import GoalStatus
+from nightingale_msgs.msg import TaskAction
+
+
+class TaskActionServer:
+    def __init__(self):
+        rospy.init_node("task_action_server_node")
+
+        self.server = actionlib.SimpleActionServer(
+            "task", TaskAction, self.goal_cb, False
```
younesr1: can the name be more descriptive?

nightingale_msgs/action/Task.action
```diff
@@ -0,0 +1,6 @@
+string name
```
younesr1: Can this "Task" idea be given a more descriptive name?

nightingale_navigation/maps/lab/jan.12.23/map.yaml
```diff
@@ -0,0 +1,7 @@
+image: map.pgm
```
younesr1: did u mean to push this

nightingale_utils/configs/room_pose_lookup.yaml
```diff
@@ -0,0 +1,45 @@
+rooms: {
```
younesr1: wasnt this part of ur other PR?

nightingale_dispatcher/src/nightingale_dispatcher/main.py
younesr1: if u make everything just 1 process, i dont this youll need this actionclient. it can just be  function call. i think that would be cleaner.

nightingale_dispatcher/src/nightingale_dispatcher/main.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self.tasks.put(goal)
+        if self.task_client.simple_state == SimpleGoalState.DONE:
+            # No goal in progress, execute goal now
+            self.task_client.send_goal(self.tasks.get(), self.handle_done_task)
+
+        return True
+
+    def handle_done_task(self, status, result):
+        if not self.tasks.empty():
```
younesr1: add logs when ending tasks and doing new tasks?

nightingale_dispatcher/src/nightingale_dispatcher/navigate_subtask.py
```diff
@@ -0,0 +1,31 @@
...
+            res = self.lookup_proxy(self.room, self.feature)
+
+            if res.status == RoomPoseLookupResponse.ROOM_NOT_FOUND:
+                return self.ARGUMENT_ERROR
+            elif res.status == RoomPoseLookupResponse.FEATURE_NOT_FOUND:
+                return self.ARGUMENT_ERROR
+            return self.SUCCESS
+
+        except rospy.ServiceException as e:
+            print(f"Room Lookup service call failed: {e}")
```
younesr1: prefer rospy log over print

nightingale_dispatcher/src/nightingale_dispatcher/navigate_subtask.py
```diff
@@ -0,0 +1,31 @@
...
+        super.__init__(name, priority)
+
+        self.room = room
+        self.feature = feature
+
+        # TODO Consider making a Singleton or share resources
+        rospy.wait_for_service("/lookup_room_pose")
+        self.lookup_proxy = rospy.ServiceProxy("/lookup_room_pose", RoomLookup)
+
+    def execute(self):
```
younesr1: is this function complete? if not can u add todo or comment or raise unimplemented exception

nightingale_dispatcher/src/nightingale_dispatcher/task_action_server.py
```diff
@@ -0,0 +1,34 @@
...
+
+import rospy
+import actionlib
+import time
+
+from actionlib_msgs.msg import GoalStatus
+from nightingale_msgs.msg import TaskAction
+
+
+class TaskActionServer:
```
younesr1: im not sure that this should be an action server. i point back to my comments about having a single process/file

nightingale_msgs/action/Task.action
younesr1: same as prev. i dont think Task.action needs to exist

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -0,0 +1,107 @@
...
+        self.room = goal.room
+        self.states.put(self.go_to_patient)
+
+        while not self.states.empty():
+            state = self.states.get()
+            state()
+        
+        self.server.set_succeeded()
+            
+def main():
```
FelixKHWong, line 115: Do we want this functionality? Shouldn't the mission planner be created on every button press?

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -0,0 +1,107 @@
...
+from nightingale_dispatcher.idle_task import IdleTask
+from nightingale_dispatcher.navigate_task import NavigateTask
+from nightingale_dispatcher.stock_task import StockTask
+from nightingale_dispatcher.task import Task
+from nightingale_dispatcher.triage_task import TriageTask
+
+class MissionPlanner:
+    def __init__(self):
+        rospy.init_node("mission_planner_node")
+
```
FelixKHWong, line 22: Do we need other variables such as the robot state and a variable (context) for what branch the robot is traversing down? I think we need robot state for sure to help update the interface on what the robot is currently doing and we can also use it to help plan other actions.

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+        self.server.start()
+
+        self.handoff_task = HandoffTask()
+        self.idle_task = IdleTask()
+        self.navigate_task = NavigateTask()
+        self.stock_task = StockTask()
+        self.triage_task = TriageTask()
+
+        self.states = queue.Queue()
```
FelixKHWong: I don't agree with using states to describe the functions here. Maybe something like "phases" would be better since its a verb or something the robot is doing. I think state should be reserved as a parameter to describe the robot on a more granular level like: "driving" or "extending arm"

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: Probably want a listener to the estop topic to be able  to listen anytime and pause things

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -0,0 +1,107 @@
...
+    
+    def go_to_patient(self):
```
FelixKHWong, line 40: I think we also need a function to go back to the home position. It seems like the idle function should be the very last thing called before the mission planner exits and waits for the next service request

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -0,0 +1,107 @@
...
+class MissionPlanner:
+    def __init__(self):
+        rospy.init_node("mission_planner_node")
+
+        self.server = actionlib.SimpleActionServer(
+            "mission_planner", MissionPlanAction, self.goal_cb, False
+        )
+        self.server.start()
+
+        self.handoff_task = HandoffTask()
```
FelixKHWong, line 32: Might also be a good idea to have a task for "arm recovery" in case the arm movement goes wrong

younesr1 (dismissed):
felix's comment are right, but im okay with this getting merged into main as it is. we can fix those issues as we go

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
le2tang: What is the expected behaviour between the proposed go_home and go_idle?

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
le2tang: Can you add this as an Issue to the repo with a description of what the task involves?

nightingale_dispatcher/src/nightingale_dispatcher/main.py
le2tang: @FelixKHWong , based on our more recent discussions, is this conversation resolved?

FelixKHWong: I think with how the new 'tasks' are structured by specifying the different actions within the function we are able to do this. This is resolved

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: I was thinking the go_home task brings the robot home from whereever it is and then queues up the go_idle task to exit. If you were already thinking of this for the idle task then this isn't necessary

le2tang: I was thinking that would go in the go_idle task, ie go_idle drives to the idle position then waits for a new task

younesr1: actually i dont think that can go in IdleTask. IdleTask doesnt have access to navigate_task so it cant call execteu. i can add this in my PR

younesr1 (approved):

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
le2tang: The IdleTask can also call room runner. Essentially for the IdleTask, the only thing to do is call room runner with the idle home room and pose

PR #24, opened 2023-01-23
Add v1 features. popups, stack refresh, nurse request

FelixKHWong:
Creating more useful features for the interface.

Added:

1. Popups to give user feedback on their selection. Popups upon arm movement.
2. Screen stack refreshing when robot finished with patient. Not critical but just for robustness
~~3. Added Nurse request screen for if patient needs physical nurse but request is still non-urgent. Important because we have taken away their nurse request button so still need method to get nurse physically in room for non-urgent requests. Currently just a screen and button but should be linked to a ROS topic or some method of alert. May need to add a companion app.~~
4. Added a start driving screen to alert people that robot will be driving. Switches to start driving screen then switches back after 5 seconds
5. Added command line parameters to allow for testing off robot and without ros more easily
6. Add useful videos to the arm retract, extend, and wait get item screens. Videos were screen captures from Rviz of the robot moving


nightingale_interface/screens/confirmation_screen.py
```diff
@@ -10,6 +14,8 @@
     UserInputs,
 )
 
+import time
```
FelixKHWong: remove

bdschnapp (dismissed):
Looks good to me! I haven't had the chance to run the code but I don't see anything that looks like it could break and I trust that it runs on your computer

nightingale-project (approved):

PR #25, opened 2023-01-29
added robot configuration lookup service

younesr1:
This PR defines some joint configurations for the robot. I found these configurations through sim and they looked good. the configurations defined will likely need to be adjusted on the robot, especially the arm configurations.

Here is a picture of the robot's home configuration:

Screenshot from 2023-01-29 16-34-54
Screenshot from 2023-01-29 16-31-36

### DIFF
--- a/nightingale_msgs/srv/RobotConfigurationLookup.srv
+++ b/nightingale_msgs/srv/RobotConfigurationLookup.srv
@@ -1,0 +1,8 @@
+# body_part is one of {left_arm,right_arm,torso,head}
+int8 LEFT_ARM=0
+int8 RIGHT_ARM=1
+int8 TORSO=2
+int8 HEAD=3
+int8 body_part
+---
+sensor_msgs/JointState jnt_state
--- a/nightingale_utils/configs/robot_configuration_lookup.yaml
+++ b/nightingale_utils/configs/robot_configuration_lookup.yaml
@@ -1,0 +1,18 @@
+robot_configurations: {
+    head: {
+        names: ["pan_joint", "tilt_joint"],
+        joints: [0.0,0.0],
+    },
+    torso: {
+        names: ["linear_joint"],
+        joints: [0.1],
+    },
+    left_arm: {
+        names: ["left_shoulder_pan_joint", "left_shoulder_lift_joint", "left_arm_half_joint", "left_elbow_joint", "left_wrist_spherical_1_joint", "left_wrist_spherical_2_joint", "left_wrist_3_joint"],
+        joints: [1.6285100157139785, 1.4853564921068, -0.4129526423290365, 2.680378098157231, 0.03587460652925101, -0.45225347904488444, 1.7006671448238166],
+    },
+    right_arm: {
+        names: ["right_shoulder_pan_joint", "right_shoulder_lift_joint", "right_arm_half_joint", "right_elbow_joint", "right_wrist_spherical_1_joint", "right_wrist_spherical_2_joint", "right_wrist_3_joint"],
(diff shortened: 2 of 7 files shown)


nightingale_msgs/srv/RobotConfigurationLookup.srv
```diff
@@ -0,0 +1,8 @@
+# body_part is one of {left_arm,right_arm,torso,head}
+int8 LEFT_ARM=0
+int8 RIGHT_ARM=1
+int8 TORSO=2
+int8 HEAD=3
+int8 body_part
+---
+sensor_msgs/JointState jnt_state
```
FelixKHWong, line 8: I think you've got a slight spelling mistake here in the variable name "int_state"

FelixKHWong (changes requested):
Just to confirm: the new service is used to query for the states of the robot so we can figure out the required motions needed to perform specific actions? So if we want to move the end effector to position X we can easily calculate the necessary cartesian movement from the current position?

nightingale_msgs/srv/RobotConfigurationLookup.srv
younesr1: jnt_state is short for joint_state, which is the type of the msg. "jnt" is a common abbreviation for "joint" used in many places in the kinova repo.

Example is line 385 in movo_jtas.py
<for jnt in joint_names>

younesr1:
> Just to confirm: the new service is used to query for the states of the robot so we can figure out the required motions needed to perform specific actions? So if we want to move the end effector to position X we can easily calculate the necessary cartesian movement from the current position?

Not quite. This PR defines some joint configurations that the robot should be in in certain situations. For example, when the robot boots up, the gimbal should face ahead and the torso should be at a predetermined height, so the mission planner will (eventually, this is not an urgent need and doesnt need to exist before saturday) query the target torso height from this server and then send a command to some other node which make the torso move.

Another example is with the arms. The service server i've created defines our own "home" configuration for the arms. So the mission-planner can query this service for the arm home configurations and then ask Ben's code to achieve these configurations.

When I say configuration i meant the specific joint angles of whatever sub-component (ie. the head). The head configuration has 2 numbers (pan & tilt)

nightingale_utils/configs/robot_configuration_lookup.yaml
```diff
@@ -0,0 +1,18 @@
+robot_configurations: {
```
le2tang: Can we change this to `joint_configurations`. I expect to have a `cartesian_configurations` that does something similar to what you have here, except in cartesian space.

nightingale_utils/src/utils.py
```diff
@@ -56,8 +58,35 @@ def room_pose_lookup(self, req):
...
+class RobotConfigurationService:
+    def __init__(self):
+        self.ss = rospy.Service(
+            "/nightingale/robot_configuration_lookup",
+            RobotConfigurationLookup,
+            self.robot_configuration_lookup,
+        )
+        self.configurations = rospy.get_param("/nightingale_utils/robot_configurations")
+
+    def robot_configuration_lookup(self, req):
```
le2tang, line 72: Nit: Add a check to make sure all the lookup arguments are defined in self.configurations

le2tang (dismissed):

nightingale_utils/configs/robot_configuration_lookup.yaml
younesr1: how about "home_joint_configurations"? these values represent "home".

bdschnapp (dismissed):

le2tang (dismissed):

FelixKHWong (dismissed):

le2tang (dismissed):

le2tang (approved):

PR #29, opened 2023-01-30
define navigate task in mission planner

younesr1:


nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
```diff
@@ -1,18 +1,30 @@
...
+                f"Failed to find the {action_name} action server",
+            )
 
-    # TODO define inputs
-    def execute(self, *args, **kwargs):
-        # Check if there are intermediate steps to travel to the destination
-        # Call room_runner to drive to pose
-        raise NotImplementedError()
+    def execute(self, room_number, sublocation):
+        goal = RoomRunnerActionGoal()
```
le2tang: Do we need to check if there are any doors in the path to get to the desired room?

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
```diff
@@ -1,18 +1,30 @@
...
+from nightingale_msgs.msg import RoomRunnerAction, RoomRunnerActionGoal
 
 
 class NavigateTask(Task):
-    # Go to position with given context
-    # TODO define parameters
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError()
+    def __init__(self):
+        action_name = "/nightingale_navigation/room_runner"
```
le2tang, line 12: Nit: Can we move this to a static constant member of NavigateTask

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -93,7 +107,7 @@ def go_idle(self):
     def goal_cb(self, goal):
         # TODO execute subtasks in order and report status
         # Nav -> Triage -> Nav -> Stock -> Nav ->
-        #   Handoff -> Idle
+        #   Handoff -> Home -> Idle
```
younesr1, line 110: The Idle task cant be in charge of driving the robot home. The object for driving the robot somewhere is tied to mission_planner and the IdleTask doesnt see it. What does the IDLE task do? should we delete it?

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
le2tang: Can this be a RoomRunnerGoal so that lines 26 is simpler: `goal.room_number = room_number`? I recall doing something similar with the arms

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
le2tang: I'm not sure what you mean by the object for driving the robot. Is that the room runner action client? If it is, can't both the navigate task and idle task classes have an instance of the client? Since the tasks are blocking, only one client will send an action goal at a time

younesr1: yeah they could but i dont think they should. i think its best for those Task types to be single instace

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
younesr1: see the comment in line 50 of the other file. if we start handling doors, we must then ask then it becomes:
nav_to_door -> door handling node -> nav_to_bedside instead of just nav to bedside

younesr1: (edit): oops i misunderstood ur comment. u r right

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
younesr1: i didnt make it a member bc its not used anywhere aside from init

nightingale_msgs/action/MissionPlan.action
```diff
@@ -2,4 +2,4 @@ string name
 ---
 int8 status
 ---
-int8 num_tasks_remaining
\ No newline at end of file
+# TODO: implement feedback int8 num_tasks_remaining
```
younesr1, line 5: @le2tang just commented out bc looks like its unused

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -38,12 +38,23 @@ def __init__(self):
         self.phases = queue.Queue()
 
     def go_to_patient(self):
-        # Extract room number and bed number from goal (bell)
-        status = self.navigate_task.execute(self.room, "bed")
+        rospy.loginfo("Nightingale Mission Planner going to patient")
+        # Assume door is open
+        # TODO: first go to doorside then bedside when door opening added
+        status = self.navigate_task.execute(self.room, "bedside")
```
le2tang, line 44: If the name is changed from `bed` to `bedside` here, the room_pose_lookup.yaml also needs to be changed

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -38,12 +38,23 @@ def __init__(self):
...
         if status == Task.ERROR:
             raise NotImplementedError()
         self.phases.put(self.triage_patient)
 
+    def go_home(self):
+        rospy.loginfo("Nightingale Mission Planner going home")
+        # Assume door is open
+        # TODO: first go to doorside then bedside when door opening added
+        status = self.navigate_task.execute("home", "")
```
le2tang, line 53: The room pose lookup service requires a name for both of the room and sublocation. The sublocation for "home" is "idle"

le2tang (changes requested):
There are some small issues with the use of the room pose lookup service, but otherwise everything looks correct

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
younesr1: ill address this once i add the lookup funcationality to room runner

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
younesr1: ill address this once i add the lookup funcationality to room runner

le2tang (approved):

PR #30, opened 2023-01-31
Add interface client to tasks

FelixKHWong:
Enable tasks to connect to interface to respond to user inputs. Also include interface constants for tasks to interpret interface results


bdschnapp (commented):
I don't understand why all tasks need this functionality. Maybe make a class that inherits from task that implements this instead?

nightingale_dispatcher/src/nightingale_dispatcher/task.py
```diff
@@ -1,6 +1,23 @@
...
 
 class Task:
-    ERROR = -1
-    SUCCESS = 0
+    def __init__(self):
+        self.ERROR = -1
+        self.SUCCESS = 0
+        self.interface_comms_service_name = BridgeConfig.UPDATE_UI_SERVICE
+
+    def interface_comms_client(self, robot_state):
```
le2tang: Can you make this function name more descriptive. Something like `wait_user_input()` or `get_user_input()` is a bit more clear about what the function is actually doing

nightingale_dispatcher/src/nightingale_dispatcher/task.py
```diff
@@ -1,6 +1,23 @@
 #!/usr/bin/env python3
+import rospy
+from nightingale_ros_bridge import BridgeConfig, RobotStatus, UserInputs
+from nightingale_msgs.srv import InterfaceCall
 
 
 class Task:
-    ERROR = -1
```
le2tang: Moving these from static variables to member variables could break things elsewhere. I remember using `Task.ERROR` as a comparison value in the mission planner PR. With this change, a Task instance would need to be constructed in order to use the constants `Task().ERROR`, which seems unnecessary

le2tang (changes requested):

nightingale_dispatcher/src/nightingale_dispatcher/task.py
younesr1: im not sure i understand. why should every child of task have this function defined? why should NavigationTask have interface_comms_client defined?

FelixKHWong: I was thinking that every task does not necessarily need to communicate with the user for input but does need to send commands to the screen to tell the screen to change screen based on what the robot is doing. for example the navigate task would send a message to the interface to change to the "driving" screen which does not get back a useful response but is still useful to change the screen. I could make a separate class that inherits from the main "task" class so only tasks that need it inherit it. Though at the moment I think each task does need to update the screen at some point

FelixKHWong: The interface communication doesn't always get user input though and sometimes is just used to update the state of the robot for the tablet to take action on it. Would "send robot state" be better?

le2tang: Ya, that name works. You know best what the function call does, so if that name is most descriptive to you, then it's good enough for me

le2tang: Ben, Felix and I had a discussion about this earlier today. We decided that many of the tasks: Stock, Triage, Handoff, all require some interface calls. It makes sense to have one place where the interface call is defined for each of these tasks, but it should not be in the Task class. The idea is to make a child class for Task called TaskWithInterface (or something like that) that has this interface client. Then the StockTask, TriageTask, and HandoffTask classes all inherit from this TaskWithInterface class instead of Task.

The inhertance structure looks like this:
class Task
class TaskWithInterface(Task)
class StockTask(TaskWithInterface)

FelixKHWong:
Increased the scope a bit to add the processing of the interface calls for the stationary tasks. Should be ok since adding hte interface calls in here

nightingale_dispatcher/src/nightingale_dispatcher/task_with_interface_comms.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+import rospy
+from nightingale_ros_bridge import BridgeConfig, RobotStatus, UserInputs
+from nightingale_msgs.srv import InterfaceCall
+from nightingale_dispatcher.task import Task
+
+
+class TaskWithInterfaceComms(Task):
+    def __init__(self):
```
le2tang, line 9: Nit: super().__init__(self) should also be called here

nightingale_dispatcher/src/nightingale_dispatcher/task_with_interface_comms.py
```diff
@@ -0,0 +1,22 @@
...
+        self.interface_comms_service_name = BridgeConfig.UPDATE_UI_SERVICE
```
le2tang: Nit: Is this variable necessary if there is a constant for it?

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -60,11 +64,13 @@ def triage_patient(self):
         status = self.triage_task.execute()
         if status == Task.ERROR:
             raise NotImplementedError()
-
-        if status == TriageTask.TIMEOUT:
+        elif status == UserInputs.WD_TIMEOUT:
+            # WD timeout
+            self.phases.put(self.go_idle)
+        elif status == UserInputs.DISMISS:
```
le2tang, line 70: Nit: Use or instead of two elif

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
```diff
@@ -28,7 +29,7 @@ def execute(self, room_number, sublocation):
         self.action_client.send_goal(goal)
         self.action_client.wait_for_result()
         return (
-            Task.SUCCESS
```
le2tang: I think all tasks should return `Task.SUCCESS` rather than the `TaskWith...` status for compatibility

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -13,7 +13,11 @@
 from nightingale_dispatcher.stock_task import StockTask
 from nightingale_dispatcher.task import Task
 from nightingale_dispatcher.triage_task import TriageTask
-from nightingale_ros_bridge.bridge_interface_config import BridgeConfig
+from nightingale_ros_bridge.bridge_interface_config import (
+    BridgeConfig,
+    RobotStatus,
+    UserInputs1,
```
FelixKHWong: will change this error

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: I was thinking we might want different handling for the two options. For now nothing will be implemented but maybe something like an alert can be posted if the watchdog times out?

le2tang (approved):

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
```diff
@@ -1,13 +1,14 @@
...
-from nightingale_dispatcher.task import Task
+
+from nightingale_dispatcher.task_with_interface_comms import TaskWithInterfaceComms
 import actionlib
 from actionlib_msgs.msg import GoalStatus
 from nightingale_msgs.msg import RoomRunnerAction, RoomRunnerGoal
 
 
-class NavigateTask(Task):
+class NavigateTask(TaskWithInterfaceComms):
```
le2tang, line 11: The NavigateTask doesn't use the interface, it can inherit directly from Task

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -60,11 +64,13 @@ def triage_patient(self):
         status = self.triage_task.execute()
         if status == Task.ERROR:
             raise NotImplementedError()
-
-        if status == TriageTask.TIMEOUT:
+        elif status == UserInputs.WD_TIMEOUT:
```
le2tang, line 67: Is there any way to standardize the outputs so that the mission planner does not depend on the constants defined in the ros bridge? Can some standard status output be given from the triage_task based on the user input

le2tang (changes requested):

nightingale_dispatcher/src/nightingale_dispatcher/navigate_task.py
FelixKHWong: I was thinking it would need to change the interface screen to the driving screen and would need to then inherit from the task with interface. Otherwise we would have to update the tablet that we are driving in a different task

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: I could duplicate the constants inside the task with interface communication task so the user inputs are evaluated and then a standardized output is returned.

FelixKHWong:
Mission planner refactored so this PR is deprecated

PR #32, opened 2023-02-01
User/Ben/Task Inheritance

bdschnapp:
Removed not implemented errors from mission planner and created a default execute method in task which raises not implemented. This is the same behavior as before, but is more generalizable and robust.

### DIFF
--- a/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
+++ b/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
@@ -42,8 +42,6 @@ class MissionPlanner:
         # Assume door is open
         # TODO: first go to doorside then bedside when door opening added
         status = self.navigate_task.execute(self.room, "bedside")
-        if status == Task.ERROR:
-            raise NotImplementedError()
         self.phases.put(self.triage_patient)
 
     def go_home(self):
@@ -51,16 +49,11 @@ class MissionPlanner:
         # Assume door is open
         # TODO: first go to doorside then bedside when door opening added
         status = self.navigate_task.execute("home", "")
-        if status == Task.ERROR:
-            raise NotImplementedError()
         self.phases.put(self.go_idle)
 
     def triage_patient(self):
         # Arrived at patient's bedside
         status = self.triage_task.execute()
-        if status == Task.ERROR:
-            raise NotImplementedError()
-
         if status == TriageTask.TIMEOUT:
             # User didn't want anything
             self.phases.put(self.go_idle)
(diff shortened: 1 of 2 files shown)


nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -114,9 +97,11 @@ def goal_cb(self, goal):
 
         while not self.phases.empty():
             phase = self.phases.get()
-            status = phase()
-
-            if not status:
+            try:
+                status = phase()
+            except Exception as e:
```
FelixKHWong, line 102: What if we want to do some specific error handling inside the tasks though? It seems like any error raised will bring it here. Are you suggesting that we should handle all of the errors in this spot? I'm concerned if we want to handle errors, have the handling succeed in bringing the robot back but then be here and unable to go back to the state it was at

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -114,9 +97,11 @@ def goal_cb(self, goal):
 
         while not self.phases.empty():
             phase = self.phases.get()
-            status = phase()
-
-            if not status:
+            try:
+                status = phase()
```
le2tang, line 101: Each phase function needs to return a status in this case

le2tang:
Overall, I think this works, but the inline NotImplementedError makes it possible to partially implement the functions, and leave the error handling for later. I think that functionality gets lost with this PR

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -42,25 +42,18 @@ def go_to_patient(self):
...
-        if status == Task.ERROR:
-            raise NotImplementedError()
         self.phases.put(self.triage_patient)
 
     def go_home(self):
         rospy.loginfo("Nightingale Mission Planner going home")
         # Assume door is open
         # TODO: first go to doorside then bedside when door opening added
         status = self.navigate_task.execute("home", "")
-        if status == Task.ERROR:
```
younesr1, line 54: keep in mind that there are currently 2 different things that are "Not Implemented". One is the execute method for some classes. The other is the behavior in case the task return FAILURE. From a quick glance at this PR it looks like ur combining both?

bdschnapp: OHHHH I completly missunderstood this, I didn't recognize that raise not implemented was being used as stubs for error handeling as well. I was assuming we were completly ignoring error checking for V1. My PR breaks your original intention for error handeling so I think this shouldn't get merged

nightingale-project:
can we close this

PR #37, opened 2023-02-01
Implement inteface communication in MP

FelixKHWong:
Implementing interface calls in all phases. 
** will need to change base to main once refactor PR is merged. Currently showing differences between refactor PR so only changed files are shown.


nightingale_dispatcher/src/nightingale_dispatcher/send_interface_request_task.py
```diff
@@ -2,12 +2,35 @@
...
-        return user_input
+        rospy.wait_for_service(BridgeConfig.UPDATE_UI_SERVICE, InterfaceCall)
+        try:
+            # spin up proxy when needed
+            self.interface_comms_proxy = rospy.ServiceProxy(
+                BridgeConfig.UPDATE_UI_SERVICE, InterfaceCall
+            )
+            user_input = self.interface_comms_proxy(robot_state)
+            # convert input to return code
+            status = Task.SUCCESS
```
FelixKHWong, line 28: looks a bit redundant but should be useful to deal with any other return data if we decide to send info to tablet

nightingale_dispatcher/src/nightingale_dispatcher/send_interface_request_task.py
```diff
@@ -2,12 +2,35 @@
...
 
 
 class SendInterfaceRequestTask(Task):
     def __init__(self):
         pass
 
     def execute(self, robot_state):
-        user_input = self.interface_comms_request()
-        return user_input
+        rospy.wait_for_service(BridgeConfig.UPDATE_UI_SERVICE, InterfaceCall)
```
younesr1: similar thing ive said before, but these wait for service can be a pain to debug. id add add rospy log statements around it (like i do in room_runner).

in the past, ive forgotten to run the service server, then my code just hangs and it comes down to these wait for service

nightingale_dispatcher/src/nightingale_dispatcher/send_interface_request_task.py
```diff
@@ -2,12 +2,35 @@
...
+            elif user_input == UserInputs.RETURN_HOME:
+                status = Task.DISMISS
+            elif user_input == UserInputs.WD_TIMEOUT:
+                status = Task.WD_TIMEOUT
+            elif user_input == UserInputs.NO_INPUT:
+                # return success
+                pass
+            return status
+        except rospy.ServiceException as e:
+            print("Service call failed: %s" % e)
```
younesr1: prefer rospy log over print always

nightingale_dispatcher/src/nightingale_dispatcher/send_interface_request_task.py
le2tang: I'm not sure this is the best way to do this. Is to use a more generic Task constant that is general enough to apply to all tasks, then map the UserInput constants to the Task constants? If not, then it's better to just use the Task constants

FelixKHWong: I'm not sure if it's possible to have a more generic task constant since the value returned has to  represent 3 or more different things. I'm thinking we either return the Task constant and also another value for the user input. Or we can have the mission planner own the constants but I'm not sure if I like that since the mission planner is just reading the values in and the ros bridge is the one who owns the communication between the interface

nightingale-project (dismissed):
Approving since my concerns are stylistic and not functional. I think you should address these comments after saturday:

return an Enum instead 1. meaning of enums is clearer
The class task should define python Enums (from enum import Enum) not the constants
nit: prefer f-strings over the % formatting u use https://www.google.com/search?q=python+fstring&oq=python+fstring&aqs=chrome..69i57j0i10i512l5j0i512j0i10i512l3.2496j0j7&sourceid=chrome&ie=UTF-8 

Felix: Adding lyndons comments to address later: I'm not sure this is the best way to do this. Is to use a more generic Task constant that is general enough to apply to all tasks, then map the UserInput constants to the Task constants? If not, then it's better to just use the Task constants

FelixKHWong:
I'll be back around 9pm to address the lint and write down the comments for a later PR. Then I will ask for reapproval and merge

le2tang (approved):

PR #38, opened 2023-02-02
add nightingale bringup pkg

younesr1:


FelixKHWong (approved):
Just wondering where are we planning on running this? on Movo2? I'm wondering if later we might want to split up what runs on which computer to make sure one computer is not overloaded. for example when perception gets more things it might need more resources

younesr1:
I was planning on running this on movo1 but we can decided where after we do some measuring of cpu load

PR #42, opened 2023-02-05
Clean up v0

FelixKHWong:
Cleaning up V0:

Things addressed:

return an Enum instead 1. meaning of enums is clearer
The class task should define python Enums (from enum import Enum) not the constants
nit: prefer f-strings over the % formatting u use https://www.google.com/search?q=python+fstring&oq=python+fstring&aqs=chrome..69i57j0i10i512l5j0i512j0i10i512l3.2496j0j7&sourceid=chrome&ie=UTF-8

Felix: Adding lyndons comments to address later: I'm not sure this is the best way to do this. Is to use a more generic Task constant that is general enough to apply to all tasks, then map the UserInput constants to the Task constants? If not, then it's better to just use the Task constants

Check and fix status conditions during each phase for mission planner

Add some necessary ros packages to rosinstall file

### DIFF
--- a/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
+++ b/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python3
-
-import rospy
-import actionlib
 import queue
 from enum import Enum
 
+import rospy
+import actionlib
 from actionlib_msgs.msg import GoalStatus
 from std_msgs.msg import String
 from nightingale_msgs.msg import MissionPlanAction
@@ -46,7 +45,7 @@ class MissionPlanner(PhaseStatus):
         # TODO: first go to doorside then bedside when door opening added
 
         # update to driving screen
-        self.send_interface_request_task.execute(RobotStatus.DRIVING)
+        task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
 
         status = self.navigate_task.execute(self.room, "bedside")
         if status == TaskCodes.ERROR:
@@ -60,7 +59,7 @@ class MissionPlanner(PhaseStatus):
         # TODO: first go to doorside then bedside when door opening added
 
         # update to driving screen
-        self.send_interface_request_task.execute(RobotStatus.DRIVING)
+        task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
 
         status = self.navigate_task.execute("home", "default")
         if status == TaskCodes.ERROR:
@@ -94,7 +93,7 @@ class MissionPlanner(PhaseStatus):
         rospy.loginfo("Nightingale Mission Planner going to stock")
 
         # update to driving screen
-        self.send_interface_request_task.execute(RobotStatus.DRIVING)
+        task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
 
         status = self.navigate_task.execute("stock", "default")
         if status == TaskCodes.ERROR:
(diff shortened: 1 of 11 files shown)


nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -1,20 +1,26 @@
...
 
 
-class MissionPlanner:
+# enum for phase status
+class PhaseStatus(Enum):
+    PHASE_COMPLETE = 1
+    PHASE_FAIL = 0
+
+
+class MissionPlanner(PhaseStatus):
```
nightingale-project: i think this inheritance is not necessary. MissionPlanner shouldnt inherit from PhaseStatus you should be able to just delete it since it looks like ur using the enum correctly (ie. PhaseStatus.PHASE_COMPLETE)

nightingale_dispatcher/src/nightingale_dispatcher/task.py
```diff
@@ -1,10 +1,17 @@
...
+    # task codes for mission planner
     ERROR = -1
     SUCCESS = 0
     STOCK_ITEMS = 1
     DELIVER_ITEMS = 2
     DISMISS = 3
     WD_TIMEOUT = 9
+
+
+class Task:
```
nightingale-project, line 15: cant we just delete this base task? i dont think its used now

nightingale_dispatcher/src/nightingale_dispatcher/task.py
```diff
@@ -1,10 +1,17 @@
 #!/usr/bin/env python3
+from enum import Enum
 
 
-class Task:
+class TaskCodes(Enum):
```
nightingale-project, line 5: nice! One thing I realized is that python allows you to compare different Enums. So TaskCodes.SUCCESS == PhaseStatus.PhaseComplete is valid and will just return False. Enums are still much more readable.

I can make a follow up PR which will prevent us from comparing different enums. so TaskCodes.SUCCESS == PhaseStatus.PhaseComplete will raise an exception that says "dont compare different enums. it makes no sense".

This should make the failure we saw saturday more clear. It wouldve resulted in a clear exception instead of the silent failure it resulted in.

nightingale_perception/package.xml
```diff
@@ -53,16 +53,22 @@
...
   <build_export_depend>robot_body_filter</build_export_depend>
   <build_export_depend>sensor_filters</build_export_depend>
+  <build_export_depend>pointcloud_to_laserscan</build_export_depend>
+  <build_export_depend>ira_laser_tools</build_export_depend>
   <exec_depend>roscpp</exec_depend>
   <exec_depend>rospy</exec_depend>
   <exec_depend>std_msgs</exec_depend>
   <exec_depend>robot_body_filter</exec_depend>
   <exec_depend>sensor_filters</exec_depend>
+  <exec_depend>pointcloud_to_laserscan</exec_depend>
```
nightingale-project, line 70: nice!

nightingale-project (dismissed):
good stuff felix! thx for keeping track of all these!

nightingale_dispatcher/src/nightingale_dispatcher/task.py
FelixKHWong: If we don't forsee having any common functionality between tasks then we can but otherwise I think it makes sense to be able to define common functionality once

FelixKHWong:
I will wait for the test day pr https://github.com/nightingale-project/nightingale/pull/43 to be merged then rebase and merge this

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: actually this might break things. I ran the mission planner but didn't get an error about inheriting the enum which is strange. I'll be removing the inheritance thanks

nightingale-project (approved):

PR #43, opened 2023-02-05
V0 test day feb 4 2023

nightingale-project:


nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -127,37 +127,37 @@ def __init__(self):
...
             return True  # TODO: action_server.get_result()
         return False
 
     def lock_right_gripper(self):
-        self.gripper_lock["right"] = 0
+        self.gripper_lock["right"] = True
 
     def unlock_right_gripper(self):
-        self.gripper_lock["right"] = 1
+        self.gripper_lock["right"] = False
```
FelixKHWong, line 154: Shouldn't 1 be true and 0 be false?

nightingale_payload/package.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>nightingale_payload</name>
```
FelixKHWong: nit: Could there be a better name for this package? I feel like payload isn't specific enough to mean the box. For communications "payload" is also used a lot for data storage and transfer. Could we maybe do like "payload_sensing"? Not the most important thing though since we all know what it means

TODO.txt
```diff
@@ -0,0 +1 @@
+the stock task is wrong bc status gets overwritten and we use the
```
FelixKHWong: I'll be addressing this in my v0 cleanup PR and a note has been made in the v0 testing data issue. Can we delete so we don't forget?

nightingale_payload/src/main.py
```diff
@@ -0,0 +1,160 @@
...
+        rospy.loginfo(f"{self.robot.link_map['right_shoulder_link'].inertial}")
+
+        self.forces = np.zeros(6)
+        self.last_forces = np.zeros(6)
+        self.filter_coeff = 0.5
+
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+
+        self.payload_pub = rospy.Publisher(
```
FelixKHWong: What exactly does this package do? I thought it senses the weight of the box and its contents. If we are basically querying the weight would a service do better here?

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
nightingale-project: i think it makes sense for true to mean that the gripper is locked. not sure why gripper locked used to be 0 but i think its good now

nightingale_msgs/CMakeLists.txt
```diff
@@ -51,11 +51,10 @@ find_package(catkin REQUIRED COMPONENTS
...
 
 ## Generate messages in the 'msg' folder
-# add_message_files(
-#   FILES
-#   Message1.msg
-#   Message2.msg
-# )
+add_message_files(
+  FILES
+  Payload.msg
```
nightingale-project: @le2tang i wont put these payload changes as part of the v0 testing PR bc they were not part of v0. feel free to make seperate PR

nightingale_payload/package.xml
nightingale-project: ill remove this pkg from this pr since its not v0

FelixKHWong (approved):

le2tang (approved):

PR #45, opened 2023-02-10
Manipulation V1

bdschnapp:
This still needs significant testing in sim and IRL, and also needs @le2tang manipulation changes to be merged in. I would appreciate an initial review focusing on architecture. Leave comments if you have any ideas on how this could break

### DIFF
--- a/nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
+++ b/nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
@@ -24,7 +24,7 @@ from tf.transformations import euler_from_quaternion
 
 # TODO: replace with service call, requires refactoring of service to include gripper info
 CFG = rospy.get_param("/nightingale_utils/joint_configurations")
-
+global manipulation
 
 # overriding the geometry msgs Pose class to add tolerance on equivalence check
 class Pose(GeometryPose):
@@ -530,31 +530,67 @@ class ManipulationControl:
 
     def home(self):
         self.right_cartesian.approximate_home()
-        print(self.jnt_ctrl.get_joint_states()["right_arm"])
         self.jnt_ctrl.home()
-        print(self.jnt_ctrl.get_joint_states()["right_arm"])
 
-    def extend_handoff(self):
-        # extend right arm in cartesian space
-        # TODO: replace with service call
-        self.jnt_ctrl.cmd_right_arm(CFG["right_arm_extended_handoff"]["joints"])
+    def extend_handoff(self, goal_point: Point):
+        point1 = Point()
+        point1.x = 0.727
+        point1.y = -0.03
+        point1.z = 0.616
+        self.right_cartesian.cmd_position(point1)
+
+        orientation1 = Quaternion()
+        orientation1.x = -0.7395762387263899
+        orientation1.y = -0.015843133434378723
+        orientation1.z = -0.07349393339288811
+        orientation1.w = 0.668860691017755
+        self.right_cartesian.cmd_orientation(orientation1)
+
+        self.right_cartesian.cmd_position(goal_point)
+
+
+def test_joint_control_home():
+    print("starting joint space home test")
+    manipulation.jnt_ctrl.home()
+    print("reached joint_control.home")
+    time.sleep(2)
+
+
+def test_cartesian_relative_movement():
+    print("starting cartesian control test")
+    pose = manipulation.right_cartesian.get_pose()
+    pose.position.x += 0.3
+    manipulation.right_cartesian.set_fixed_ee_ctrl_mode()
+    manipulation.right_cartesian.cmd_arm(pose)
+    print("reached position 1")
+    time.sleep(5)
+
+
+def test_cartesian_homing_sequence():
+    print("starting cartesian home sequence test")
+    manipulation.home()
+    print("reached manipulation.home")
+    time.sleep(5)
+
+
+def test_extend_handoff():
+    print("starting extend handoff test")
+    point = Point()
+    point.x = 0.7
+    point.z = 0.5
+    manipulation.extend_handoff(point)
+    print("finished extend handoff test")
+    time.sleep(5)
 
 
 # test code
 if __name__ == "__main__":
     rospy.init_node("manipulation_control")
+    global manipulation
     manipulation = ManipulationControl()
 
-    manipulation.home()
-    print("reached home")
-    time.sleep(2)
-
-    pose = manipulation.right_cartesian.get_pose()
-    print(pose)
-
-    pose.position.x += 0.3
-
-    manipulation.right_cartesian.set_fixed_ee_ctrl_mode()
-    print(manipulation.right_cartesian.cmd_arm(pose))
-
-    print(manipulation.right_cartesian.get_pose())
+    test_joint_control_home()
+    test_cartesian_relative_movement()
+    test_cartesian_homing_sequence()
+    test_extend_handoff()
+    test_joint_control_home()
--- a/nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
+++ b/nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
@@ -79,11 +79,7 @@ class Orientation:
 
 
 def joint_goal(
-        joint_values: list,
-        joint_names: list,
-        eev=0.5,
-        eea=0.5,
-        timeout=5
+    joint_values: list, joint_names: list, eev=0.5, eea=0.5, timeout=5
 ) -> MoveToJointsMoveItGoal:
     goal = MoveToJointsMoveItGoal()
     goal.endEffectorVelocity = eev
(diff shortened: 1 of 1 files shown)


nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -2,29 +2,97 @@
...
+        return all(
+            [
+                math.isclose(self.roll, other.roll, rel_tol=self.tol),
+                math.isclose(self.pitch, other.pitch, rel_tol=self.tol),
+                math.isclose(self.yaw, other.yaw, rel_tol=self.tol),
+            ]
+        )
+
+
+def quaternion_to_euler(q: Quaternion):
```
le2tang: There is a function in tf_conversions to do this already

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -76,19 +148,129 @@ def __init__(self):
...
+        self._left_joint_states = None
+        self._joint_tolerance = joint_tolerance
 
-    def cmd_left_arm(self, joint_values, blocking=True):
-        goal = joint_goal(joint_values, self.left_arm_joint_names)
-        self.left_arm.send_goal(goal)
-        if blocking:
-            return self.left_arm.wait_for_result()  # TODO: action_server.get_result()
-        return True
+    def update_joint_states(self):
```
le2tang, line 142: Why not use a subscriber to always get the most recent data, then store it and use it as necessary, rather than having a function that only listens for 1 message

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -76,19 +148,129 @@ def __init__(self):
...
+
+    def cmd_right_arm(self, joint_target: list, blocking=True) -> bool:
+        """
+        Commands movement of the right arm through the right_arm action server
+
+        @param joint_target: list of target joint values
+        @param blocking: default True, if blocking, function will wait for action server response
+        @return: result of verify_joint_target and action server state if blocking
+        """
+        if self.verify_joint_target(joint_target, self._right_joint_states):
```
le2tang: Nit: Instead of nesting if statements, use early returns with inverted conditions
`if not {condition1} return\n
if not {condition2} return\n
goal  = ...\n
send_goal()`

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
bdschnapp: I spent so long looking for an existing function -_-

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
bdschnapp: I believe doing it this way increases performance (and possibly reduces network traffic?), espicially when the manipulation task isn't running. The update is only needed before a control command is sent so it doesn't really make sense to me to continiously update

bdschnapp:
This still needs to be merged with lyndon's robot pose lookup service changes but otherwise it is ready for review

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -76,26 +135,134 @@ def __init__(self):
...
+
+        except rospy.ROSException:
+            rospy.logwarn(
+                "Unable to get data from /joint_states, wait_for_message timeout"
+            )
+            return False
+
+        names_pos_dict = {}
+        for i in range(len(names)):
+            names_pos_dict[names[i]] = pos[i]
```
younesr1, line 163:
```suggestion
        for n,p in zip(names,pos):
            names_pos_dict[n] = p
```

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -2,30 +2,85 @@
...
+from tf.transformations import euler_from_quaternion
 
 # TODO: replace with service call, requires refactoring of service to include gripper info
 CFG = rospy.get_param("/nightingale_utils/joint_configurations")
 
 
-def joint_goal(joint_values, joint_names, eev=0.5, eea=0.5, timeout=10):
+# overriding the geometry msgs Pose class to add tolerance on equivalence check
+class Pose(GeometryPose):
+    def __init__(self, tolerance=0.01):
```
younesr1, line 31: hmm 0.01 rad seems pretty small. its half a degree. is the encoder resolution smaller than this? if not this wont do much i dont think

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -76,26 +135,134 @@ def __init__(self):
...
+        @return: True if any targets are different from the current state,
+                 False if all targets are the same as current state
+        """
+
+        if len(joint_target) != 7:
+            raise Exception("Invalid joint value length")
+        error_count = 7
+        for i in range(7):
+            if math.isclose(
+                abs(joint_target[i]),
```
younesr1, line 209: why take the abs value? also you can get rid of the for loop with https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -76,26 +135,134 @@ def __init__(self):
...
-    def home(self):
-        self.cmd_left_arm(self.left_arm_home_joint_values, blocking=False)
         self.left_arm.wait_for_result()
-        self.cmd_right_arm(self.right_arm_home_joint_values, blocking=False)
-        self.right_arm.wait_for_result()
-        return True  # TODO: action_server.get_result()
+        status = self.left_arm.get_result().status
+        if status == "Success":
+            return True
+        rospy.logwarn("cmd_left_arm failed with status: " + str(status))
```
younesr1, line 259: nit: log error

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +299,215 @@ def __init__(self):
...
         )
-        self.right_arm.wait_for_server()
+        self.arm.wait_for_server()
+        self.prefix = prefix
+        self.ee_ctrl_mode = 0
+        self.orientation = Orientation()
+        self.pose = Pose()
+        self.ref_link = "base_link"
+        self.ee_link = f"{prefix}_ee_link"
+        # TODO: include correct home pose
```
younesr1, line 355: will this PR add homing

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +299,215 @@ def __init__(self):
...
+                pose.orientation.x,
+                pose.orientation.y,
+                pose.orientation.z,
+                pose.orientation.w,
+            ]
+            (roll, pitch, yaw) = euler_from_quaternion(quaternion_list)
+            euler_orientation = Orientation(roll=roll, pitch=pitch, yaw=yaw)
+            goal = cartesian_goal(pose.position, ref_link, euler_orientation)
+
+        print("goal: ")
```
younesr1, line 432: id prefer rospy logging over print

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +299,215 @@ def __init__(self):
...
+        print(goal)
+
+        self.arm.send_goal(goal)
+        if blocking:
+            self.arm.wait_for_result()
+            return self.arm.get_result()
```
younesr1, line 438: can this return a bool? so check if get_result() is good before u return? the caller wont know what the meaning of the get_result() is (unless get_result() already gives u a bool)

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -190,42 +517,22 @@ def __init__(self):
...
-        """
-        raise NotImplementedError()
+        rospy.loginfo("Manipulation initialization successful")
 
     def home(self):
-        # home left arm in joint space and home right arm in cart or joint space
-        # use self.jnt_ctrl.home() for now
-        raise NotImplementedError()
+        self.right_cartesian.approximate_home()
+        print(self.jnt_ctrl.get_joint_states()["right_arm"])
```
younesr1: id use rospy logs over prints

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +299,215 @@ def __init__(self):
         self.unlock_right_gripper()
 
     def cmd_right_gripper(self, goal, blocking=True):
```
younesr1, line 301: this function still has a blocking var

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +299,215 @@ def __init__(self):
...
+        self.orientation.roll = roll
+        self.orientation.pitch = pitch
+        self.orientation.yaw = yaw
+
+        p = link_transform.transform.translation
+        self.pose.position = p
+        self.pose.orientation = q
+        return True
+
+    def cmd_arm(self, pose: GeometryPose, blocking=True, ee_relative=False):
```
younesr1, line 395: i thought we were remiving blocking

younesr1:
i think we should close this in favor of : https://github.com/nightingale-project/nightingale/pull/53/files

PR #46, opened 2023-02-11
Remove arm init sim

FelixKHWong:
Removed the timed launch of the arm init procedure since it interferes with the planning with the box. The launch file call I have replace only performs the arm init so no nightingale movo  functionality is lost.

Example movement with box after arm init removed. Arms start out in front but eventually mission planner will home the arms upon startup
box_movement


younesr1 (commented):
we shouldnt merge this in before the Nightingale homing sequence is added. the robot cannot drive with the arms out. it will tip over. this prob means we  dont want to merge the addition of the box until after the nightingale homing sequence is added.

bdschnapp (approved):

PR #47, opened 2023-02-17
reduce inflatino when close to goal + more features

younesr1:
This PR does a couple of things:
1. This PR reduces the costmap inflation when the robot is < 2.5 meters away from the goal and inflates it back up when its more than 7 meters from the goal. this allows the robot to get up real close to the bed.
2. This PR increases the planner patience. The planner used to wait 5 seconds before beginning space clearing behaviors. it now waits 15 seconds. 
3. it increase sthe shape of the polygon the base uses during nav bc we now drive with a box
4. it moves rotational space clearing behavior until after an aggressive reset of the costmap is performed. the rotational clearing is desperate and should be done after the aggressive map reset. i also added another aggressive map reset after the rotational clearing
5. we now clear the costmaps on every navigation request. on thursday, i noticed that the robot would last see someone behind a hallway before it makes a turn and never sees them leave. so it assumes that person is still there which causes it to be unable to return home. clearing the costmap should leave only static obstacles and currently observable obstacles (in theory. i will test later)

### DIFF
--- a/nightingale_navigation/configs/costmap_common_params.yaml
+++ b/nightingale_navigation/configs/costmap_common_params.yaml
@@ -6,7 +6,8 @@ robot_base_frame:           base_link
 transform_tolerance:        0.5
 
 # ---Vector footprint (Meters):---
-footprint: [[0.39,0.3],[0.39,-0.3],[-0.39,-0.3],[-0.39,0.3]]
+#footprint: [[0.39,0.3],[0.39,-0.3],[-0.39,-0.3],[-0.39,0.3]]
+footprint: [[0.87,0.52],[0.87,-0.4],[-0.56,-0.4],[-0.56,0.52]]
 
 ### Costmap definitions ###
 update_frequency: 8.0 # laserscans are 15 Hz. Be conservative
@@ -15,5 +16,5 @@ inflation_radius: 0.45
 obstacle_layer/footprint_clearing_enabled: true
 
 #raytrace_range must be less than max range of sensor.
-observation_sources: scan_multi
-scan_multi: {data_type: LaserScan, topic: movo/scan_multi, marking: true, clearing: true, inf_is_valid: true, obstacle_range: 4.75, raytrace_range: 8.0}
+observation_sources: base_scan_filtered
+base_scan_filtered: {data_type: LaserScan, topic: movo/base_scan_filtered, marking: true, clearing: true, inf_is_valid: true, obstacle_range: 3.5, raytrace_range: 8.0}
--- a/nightingale_navigation/src/room_runner.py
+++ b/nightingale_navigation/src/room_runner.py
@@ -14,6 +14,8 @@ import tf2_ros
 import tf_conversions
 from actionlib_msgs.msg import GoalStatus
 from geometry_msgs.msg import PoseWithCovarianceStamped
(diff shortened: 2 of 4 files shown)


nightingale_navigation/configs/costmap_common_params.yaml
```diff
@@ -15,5 +16,5 @@ inflation_radius: 0.45
 obstacle_layer/footprint_clearing_enabled: true
 
 #raytrace_range must be less than max range of sensor.
-observation_sources: scan_multi
-scan_multi: {data_type: LaserScan, topic: movo/scan_multi, marking: true, clearing: true, inf_is_valid: true, obstacle_range: 4.75, raytrace_range: 8.0}
+observation_sources: base_scan_filtered
```
FelixKHWong, line 19: What was the decision to change to base_scan_filtered?

FelixKHWong (dismissed):
looks good. I'm assuming the behaviour was testing on Thursday because we were able to see the costmap clear and also get close to the bed?

nightingale_navigation/configs/costmap_common_params.yaml
younesr1: the laser scanner kept reporting a flickering obstacle in the hallway when the robot was in the room. the obstacle was not real bc the robot was in the lab and cannot see into the hallway. If i had to guess, id say the culprit is the reflective wheel on the desk the robohub folks just added to the room. the presence of this fake flickering point caused us to take the "long way around" to the stocking area. it was transient so didnt happen every time. 

anyway, the filtered scan rejects this lone, flickering point.

younesr1: the flickering point exists in the unfiltered scan, but does not exist in the filtered scan

nightingale_navigation/src/room_runner.py
```diff
@@ -147,6 +191,14 @@ def fb_cb(fb):
                 [fb.base_position.pose.position.x, fb.base_position.pose.position.y]
             )
             euclidean_distance = np.linalg.norm(target - cur)
+            if euclidean_distance < self.deflation_hysteresis_low:
+                self.dynamic_reconfigure_client.update_configuration(
+                    {"inflation_radius": self.local_inflation / 3}
```
le2tang, line 196: Maybe add a parameter to control the deflation factor

nightingale_navigation/src/room_runner.py
```diff
@@ -147,6 +191,14 @@ def fb_cb(fb):
                 [fb.base_position.pose.position.x, fb.base_position.pose.position.y]
             )
             euclidean_distance = np.linalg.norm(target - cur)
+            if euclidean_distance < self.deflation_hysteresis_low:
```
le2tang, line 194: I'm not sure this line guarantees hysteresis. Is hysteresis the desired behaviour? It seems like this method might work fine

le2tang (dismissed):

nightingale_navigation/src/room_runner.py
nightingale-project: im not sure what u mean?

nightingale-project: yes, hysteresis is desired

PR #49, opened 2023-02-19
User/felix/pose estimation node

FelixKHWong:
Currently a work in progress but coming soon.

Currently takes the latest image, performs pose extraction, calculates the plane of the person and extrapolates a position for the bin to be set at. The bin goal is adjusted to be 10cm upwards from the center of the chest to account for the difference between the bin and the end effector so it is more comfortable to interact with the bin.

Currently requires a good initial image otherwise the position estimate does not have any error checking and will execute based on the given image. 1 method is to be at a position in which the body is fully visible and then call the action to execute the process.

Runtime is very fast.

Action returns a Pointstamped message with a single coordinate in the base id frame.

Result of testing
perception_nav_tested
The goal is slightly off to the left but is at the expected height. This may be due to noise in the depth image causing the shoulder plane to be on a slight angle. The points shown are:
1. furthest out point is the bin goal
2. The point on the face is the right eye point
3. The shoulder point on the right shoulder
4. Not shown is the left shoulder point since it is in the body
5. The point above the shoulder is the synthetic point created to get a vertical plane for better bin placement
The goal is compared with the arm to see if it is reachable. This position is determined using the ratio of the shoulders to the arms. Will likely be adjusted to make the position closer since it is slightly far from the body here

### DIFF
--- a/nightingale_bringup/launch/nightingale_system.launch
+++ b/nightingale_bringup/launch/nightingale_system.launch
@@ -4,6 +4,7 @@
     <!-- This file launches all launch files performing core nightingale functions -->
     <include file="$(find nightingale_utils)/launch/utils.launch"/>
     <include file="$(find nightingale_perception)/launch/pt_cloud_process.launch"/>
+    <include file="$(find nightingale_perception)/launch/pose_estimation.launch"/>
     <include file="$(find nightingale_navigation)/launch/navigation.launch"/>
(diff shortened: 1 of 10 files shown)


FelixKHWong:
This PR has been merged into the V1-test day branch and will be closed soon

FelixKHWong:
Deprecated for https://github.com/nightingale-project/nightingale/pull/57 since this new pr has mission planner integration and error handling

PR #50, opened 2023-02-20
added hallway planner

younesr1:


nightingale_navigation/hallway_planner/include/hallway_planner/hallway_planner.hpp
```diff
@@ -0,0 +1,135 @@
+#pragma once
```
younesr1: ill seperate implementation and definition later

nightingale_navigation/hallway_planner/src/hallway_locator.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <hallway_planner/hallway_locator.hpp>
+#include <opencv2/imgproc.hpp>
+#include <ros/ros.h>
+#include <numeric>
+
+
+namespace hallway_planner::hallway_locator
+{
+
+std::optional<cv::Mat> locate_hallways(const costmap_2d::Costmap2DROS *costmap, const HallwayLocatorConfig &config)
```
FelixKHWong: would it be useful to split these two functions with suffixes of "costmap_ros" and "costmap_cv"? It looks a bit confusing to be calling the same function that's overloaded.

nightingale_navigation/hallway_planner/src/geometry_primitives.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+double Line::length() const
+{
+    return euclidean_distance(a_,b_);
+}
+
+double Line::calculate_slope(const Point &a, const Point &b)
+{
+    // let slope be positive since choosing which point is first is arbitrary
+    const double num = a.y()-b.y();
```
FelixKHWong: just curious and not feedback but why use double over auto here?

nightingale_navigation/hallway_planner/src/hallway_planner.cpp
```diff
@@ -0,0 +1,8 @@
+/*********************************************************************
+ * Software License Agreement (MIT License)
+ * Author: Younes Reda
+ *********************************************************************/
+#include <pluginlib/class_list_macros.hpp>
+#include <hallway_planner/hallway_planner.hpp>
+
+PLUGINLIB_EXPORT_CLASS(hallway_planner::HallwayPlanner, nav_core::BaseGlobalPlanner)
```
FelixKHWong: could you send the documentation on how the plugin stuff works? I don't think it's worth you explaining it to me so I can read up on it

nightingale_navigation/hallway_planner/src/hallway_locator.cpp
```diff
@@ -0,0 +1,137 @@
...
+                ret.emplace_back(cur,*most_similar);
+                filtered_lines.erase(most_similar);
+                filtered_lines.pop_front();
+            }
+        }
+        return ret;
+    }();
+    ROS_INFO_STREAM("hallway_planner::hallway_locator::locate_hallways found " << hallways.size() << " hallways."); 
+    cv::Mat hallway_map(costmap.size(),costmap.type(),HallwayLocatorConfig::NO_HALLWAY);
+    for (const auto &hallway : hallways)
```
FelixKHWong: Is this where you will assign the costs inside the hallway? So far it seems like you are just creating the hallway space correct?

nightingale_navigation/hallway_planner/include/hallway_planner/hallway_planner.hpp
```diff
@@ -0,0 +1,135 @@
...
+       * @param plan The plan... filled by the planner
+       * @param cost The plans calculated cost
+       * @return True if a valid plan was found, false otherwise
+       */
+      bool makePlan(const geometry_msgs::PoseStamped& start, 
+                    const geometry_msgs::PoseStamped& goal, std::vector<geometry_msgs::PoseStamped>& plan,
+                    double& cost) override
+      {
+        cost = -1;
+        if (hallway_planner::HallwayPlanner::makePlan(start,goal,plan))
```
FelixKHWong: how will the plan be made? Will you be running an existing version of A* on the costmap you generated from the hallway locator?

FelixKHWong (dismissed):
There seems to be a lot going on here which I don't have the background knowledge of to evaluate that well though from the demo you showed us of the hallway locator a while back I can get a grasp of what is happening. My c++ is also not that strong so I don't have much skill to help evaluate what you've done and if there are any code practices that are not good. I'm going to go with it if it works it works since we are coming short on time

nightingale_navigation/hallway_planner/src/geometry_primitives.cpp
younesr1: the y() function returns an int. if i use auto, num will be an int. when i do division, it will be integer division which would be wrong

nightingale_navigation/hallway_planner/src/hallway_planner.cpp
younesr1: i kinda just copied the globalplanner code and defined my own plugin but this link says more http://wiki.ros.org/pluginlib

nightingale_navigation/hallway_planner/src/hallway_locator.cpp
younesr1: yeah so this function returns a a bitwise matrix. if theres a 255, then it indicates that that cell is part of a hallway. if theres a 0, then it indicates that cell is not part of a hallway

nightingale_navigation/hallway_planner/include/hallway_planner/hallway_planner.hpp
younesr1: the original plan will be made through using the old GlobalPlanner (the default globalplanner that ROS provides) which does A* or djikstra depending how u config. i think we config djikstra. Then the hallways are detected and the path is biased

bdschnapp (approved):

PR #53, opened 2023-02-22
User/yreda/arm surgery

younesr1:


nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +295,216 @@ def __init__(self):
...
+        if not self.gripper_lock["left"]:
+            return False
+
+        self.left_gripper.send_goal(goal)
+        self.left_gripper.wait_for_result()
+        return self.left_gripper.get_result().status == "Success"
 
     def lock_right_gripper(self):
-        self.gripper_lock["right"] = True
+        self.gripper_lock["right"] = False
```
FelixKHWong, line 314: Are these booleans correct? It feels like if we are locking the gripper it should be True and vice versa

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -132,55 +295,216 @@ def __init__(self):
...
+        rospy.loginfo("goal: ")
+        rospy.loginfo(goal)
+
+        self.arm.send_goal(goal)
+        if blocking:
+            self.arm.wait_for_result()
+            return self.arm.get_result().status == "Success"
+        return True
+
+    def cmd_position(self, point: Point, ee_fixed=True):
```
FelixKHWong, line 438: Should "Point" be a point stamped so it has the frame of the point? Or is it always assumed to be in the base_link frame? If this remains as just a point it is ok we just need to extract out the contents of the PointStamped() provided by the perception node which I can do easily.

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -190,51 +514,96 @@ def __init__(self):
...
-        # home left arm in joint space and home right arm in cart or joint space
-        # use self.jnt_ctrl.home() for now
-        raise NotImplementedError()
+        # CAUTION: This function should only ever home the arms. Don't add homing of other things here
+        # right gripper is openend on bootup by kinova. not sure where, but not in init
+        if not self.gpr_ctrl.close_right():
+            return False
+        if not self.jnt_ctrl.cmd_left_arm(self.jnt_ctrl.left_arm_home_joint_values):
+            return False
+        home_pose = GeometryPose()
```
FelixKHWong, line 535: What joint does "home_pose" refer to? Is it always the end effector link of the arm? What if we want to adjust each joint to a specific position so we avoid any "flaring" of the arms. this can work for now but might be an issue later if the arm flaring is more of an issue

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -190,51 +514,96 @@ def __init__(self):
...
+    handoff_pose.orientation.x = -0.69435
+    handoff_pose.orientation.y = -0.05705
+    handoff_pose.orientation.z = -0.054539
+    handoff_pose.orientation.w = 0.7153
+
+    rospy.loginfo("Going to home orientation")
+    manipulation.right_cartesian.cmd_orientation(home_pose.orientation)
+    rospy.loginfo("Arrived at home orientation")
+
+    for iter in range(3):
```
FelixKHWong, line 598: Should this be here? Not sure if this is going to mess with the testing when the node is brought up

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
younesr1: i agree its logically backwards, but it should still work. you can bring this up in the V1 PR

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
younesr1: true this shuold be a pointstamped with the frame. dont know that this super important for today

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
younesr1: in this context (the manipulators context since this is the manipulators pkg) home refers to the homing of the arms. we home left arm in joint space, but home the right arm in cartesian. this is bc the orientation lock only works for cartesian (as ive been told). we shouldnt command each joint on the right arm bc itd flip the box

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
younesr1: this chunk of code only runs if we rosrun this node. during our full runs, we do not rosrun this node. file just defines some classes used elsewhere

le2tang (approved):

younesr1:
Closing this PR bc the v1 test day PR includes these changes and prob more fixes

PR #54, opened 2023-02-23
User/felix/arm cartesian lookup

FelixKHWong:
The start of the cartesian lookup service so common states such as home and extend for stock can be looked up. 

Added to the existing lookup by adding parameters of "cart_state" and "joint_name". User will set cart_state to true if using cartesian and must additionally specify the joint_name so the corresponding position for that action can found. The lookup works by having an array of poses(xyz,xyzw) and the code finds the corresponding index of the joint name to be commanded which returns an array of 7 values describing the pose the joint should be at.

Just a start and I'm hoping someone with more knowledge on how the service works can take it and test it.

May need to be rebased on main after the updated to the current lookup service is merged


nightingale_utils/configs/robot_configuration_lookup.yaml
```diff
@@ -3,40 +3,101 @@ joint_configurations: {
...
             joints: [0.1],
+            cartesian: [
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+            ],
         },
         left_arm: {
             names: ["left_shoulder_pan_joint", "left_shoulder_lift_joint", "left_arm_half_joint", "left_elbow_joint", "left_wrist_spherical_1_joint", "left_wrist_spherical_2_joint", "left_wrist_3_joint"],
             joints: [1.6285100157139785, 1.4853564921068, -0.4129526423290365, 2.680378098157231, 0.03587460652925101, -0.45225347904488444, 1.7006671448238166],
+            cartesian: [
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
```
le2tang, line 22: There should be a position and orientation (quaternion) array here instead of a matrix.

nightingale_utils/src/utils.py
```diff
@@ -69,25 +69,76 @@ def __init__(self):
...
-        configuration = self.configurations[req.configuration]
-
-        if body_part_string not in configuration.keys():
-            return None
-
-        ret.jnt_state.name = configuration[body_part_string]["names"]
-        ret.jnt_state.position = configuration[body_part_string]["joints"]
-        ret.jnt_state.velocity = [0.0 for jnt in ret.jnt_state.name]
-        return ret
+        if req.lookup_cart == True:
```
bdschnapp, line 72: I would suggest removing req.lookup_cart. its a lot easier to just fill the jnt_state and cart_state response values and let the user decide what to do with the response rather than requiring them to declare in the request which values they need

nightingale_msgs/srv/RobotConfigurationLookup.srv
```diff
@@ -5,5 +5,8 @@ int8 RIGHT_ARM=1
 int8 TORSO=2
 int8 HEAD=3
 int8 body_part
+bool lookup_cart
+string joint_name
```
bdschnapp, line 9: all cartesian poses are from base_link -> body_part end effector, so I believe the joint name can be removed.

bdschnapp (changes requested):

PR #55, opened 2023-02-23
V1 Test Day

younesr1:


nightingale_utils/configs/robot_configuration_lookup.yaml
```diff
@@ -10,7 +10,8 @@ joint_configurations: {
         },
         left_arm: {
             names: ["left_shoulder_pan_joint", "left_shoulder_lift_joint", "left_arm_half_joint", "left_elbow_joint", "left_wrist_spherical_1_joint", "left_wrist_spherical_2_joint", "left_wrist_3_joint"],
-            joints: [1.6285100157139785, 1.4853564921068, -0.4129526423290365, 2.680378098157231, 0.03587460652925101, -0.45225347904488444, 1.7006671448238166],
+            joints: [1.4463922159520797, 1.596695035763699, 0.9848071437070933, 2.5494744715692814, -2.004909960289334, 1.8316609450514283, -2.8256088565680004],
```
younesr1, line 13: this file shouldnt have home::left_arm and left_arm_home. i think this was a bug that got merged in somehow. we should have a single concept and defintion of arm home.

- [ ] Fix this

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -130,110 +296,362 @@ def __init__(self):
...
-        raise NotImplementedError()
+    def extend_handoff(self, goal_point: Point):
+        def extend_handoff_internal():
+            return self.right_cartesian.cmd_position(goal_point, True)
 
-    def home(self):
-        # home left arm in joint space and home right arm in cart or joint space
-        # use self.jnt_ctrl.home() for now
-        raise NotImplementedError()
+        for _ in range(self.tries):
```
younesr1: - [x] tries should be passed into the function call and default to 3

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -130,110 +296,362 @@ def __init__(self):
...
+        return False
 
 
 # test code
 if __name__ == "__main__":
     rospy.init_node("manipulation_control")
     manipulation = ManipulationControl()
-    manipulation.jnt_ctrl.home()
+
+    '''# close the gripper
```
younesr1: - [x] delete this commented out hcunk

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -130,110 +296,362 @@ def __init__(self):
...
+
+        time.sleep(2)
+
+        rospy.loginfo("Going to handoff pose")
+        manipulation.right_cartesian.cmd_position(handoff_pose.position, True)
+        rospy.loginfo("Arrived at handoff pose")
+
+        time.sleep(2)'''
+
+    print(manipulation.jnt_ctrl.left_arm_home_joint_values)
```
younesr1: - [x] rospy over print

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -2,29 +2,85 @@
...
 
+# Kinematics
+from sensor_msgs.msg import JointState
+import tf_conversions, tf2_ros
+from geometry_msgs.msg import Pose as GeometryPose
+from geometry_msgs.msg import Quaternion, Point
+from tf.transformations import euler_from_quaternion
 
 # TODO: replace with service call, requires refactoring of service to include gripper info
+CFG = rospy.get_param("/nightingale_utils/joint_configurations")
```
younesr1: - [x] this shoud be deleted and use the service instead

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
```diff
@@ -75,26 +135,132 @@ def __init__(self):
...
+            return False
+        if not self.verify_joint_target(joint_target, self._right_joint_states):
+            return True
+
+        goal = joint_goal(joint_target, self.right_arm_joint_names)
+        self.right_arm.send_goal(goal)
+
+        self.right_arm.wait_for_result()
+        status = self.right_arm.get_result().status
+        if status == "Success":
```
younesr1: - [x] avoid comparison to string literals like this a typo can be a silent bug. define a var and use it instead of the literal

FelixKHWong:
Arm fixes todo:
1. Remove reorienting and closing the gripper on every movement
2. Arm seems to move in weird ways at times
3. Arm trajectory takes a long time to plan
4. Possible high CPU usage slowing down planning?
5. Changed arm home
6. Fix lookup of joint states/cartesian states
7. add parameter to specify the timeout of the arm planning
~~8. Ensure that moveit is using camera to detect obstacles for arm planning~~
9. @@@@@ found that arm planning doesn't care about obstacles in the way, only obstacles in the final position. Arm will proceed to move through obstacles if they are not obstructing final pose
Nav fixes:
1. change polygon to accomodate new arm home(87cm front, 56cm back, 52cm left, 40cm right)
2.make costmap inflate again when not near goal

Added retrial of arm planning so that a higher likelihood of success is found.

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -150,36 +157,38 @@ def handoff_items_phase(self):
         rospy.loginfo("Nightingale Mission Planner starting to hand items")
         # Arrived at patient's bedside
 
+        # pose estimation
```
younesr1: were leaving the pose estimation commented out for now to make arm testing easer

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
younesr1: or maybe the caller should use the for loop

nightingale_dispatcher/src/nightingale_dispatcher/move_arm_task.py
```diff
@@ -9,8 +9,20 @@
...
-        assert len(end_pose) is 7
-        self.manipulation.jnt_ctrl.cmd_right_arm(end_pose)
-        return TaskCodes.SUCCESS
+    def extend_restock(self):
+        result = self.manipulation.extend_restock()
+        assert type(result) is bool
+        return TaskCodes.SUCCESS if result else TaskCodes.ERROR
+
+    def retract_right_arm(self):
+        result = self.manipulation.home_right()
```
younesr1: - [x] home_right() currently closes the right gripper. this function should just retract the right arm back to home

nightingale_dispatcher/src/nightingale_dispatcher/move_arm_task.py
```diff
@@ -9,8 +9,20 @@
 class MoveArmTask(Task):
     def __init__(self):
         self.manipulation = ManipulationControl()
+        if not self.manipulation.home_left() or not self.manipulation.home_right():
```
younesr1, line 12: - [x] home_left should close the left gripper. the left gripper is opened by kinova

nightingale_msgs/action/PoseEstimation.action
```diff
@@ -0,0 +1,10 @@
+# input is the body part as a string
+# {"body", "head"}
+string target
+---
+# returns a 3D coordinate PointStamped
+# of the bin goal in base link frame
+geometry_msgs/PointStamped bin_goal
+---
+# returns if issue getting pose or accurate coords 
+int8 pose_detect_failed
```
younesr1: @FelixKHWong  this is the feedback part of ur action. not the return part. the idea behind feedback, is that its continuously published. so for RR, the feedback is the distance to the goal. we constantly send that as feedback. 

do u continuous publish whether the pose_detected_failed while the aciton is running? if not, then this should be part of the result.

nightingale_navigation/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ catkin_package(
 ## in contrast to setup.py, you can choose the destination
 catkin_install_python(PROGRAMS
   src/room_runner.py
+  src/test_do_not_merge.py
```
younesr1: navigation related changes will be removed from this PR and merged in a different PR

nightingale_utils/configs/robot_configuration_lookup.yaml
```diff
@@ -31,7 +32,7 @@ joint_configurations: {
 
     left_arm_home: {
         names: ["left_shoulder_pan_joint", "left_shoulder_lift_joint", "left_arm_half_joint", "left_elbow_joint", "left_wrist_spherical_1_joint", "left_wrist_spherical_2_joint", "left_wrist_3_joint" ],
-        joints: [1.5331797090297457, 1.605558035871483, 0.722653222587827, 2.606298339748302, -2.3331829141612968, 1.825435539687259, -2.9605818553548513]
+        joints: [1.4463922159520797, 1.596695035763699, 0.9848071437070933, 2.5494744715692814, -2.004909960289334, 1.8316609450514283, -2.8256088565680004],
```
younesr1: we had to redefine a new home for the left arm bc it was too close to the body

nightingale_navigation/src/test_do_not_merge.py
```diff
@@ -0,0 +1,17 @@
+from nightingale_dispatcher.navigate_task import NavigateTask
```
FelixKHWong: careful not to merge

nightingale_msgs/action/PoseEstimation.action
FelixKHWong: oh I see. Since the node returns a lot quicker than I thought(could be a service) I'll remove this and we will just treat the perception node as a service(even though it's an action) because it is already implemented as such

nightingale_utils/configs/robot_configuration_lookup.yaml
bdschnapp: This is temporary code and will be fixed by lyndon's lookup service change

nightingale_manipulation/src/nightingale_manipulation/manipulation_action_client.py
bdschnapp: done in lyndons pr

nightingale-project (approved):
checked diff locally

PR #56, opened 2023-02-26
Add error handling for pose estimation

FelixKHWong:
Adding error handling to pose estimation in case of perception failure. The pose estimation task will try to get the pose up to 3 times and will fail if this does not happen.

Behaviour is to go to a predefined position which is determined to likely be safe(currently set to the restock position).

For the time being, we should aim to ensure that there is always a good pose given to the robot(make sure a person is in a good position which shoulders clearly in view). This functionality should be important but with our demo being a video and the limited time, the error handling is not critical.

Merging int v1 test day so we have all of the code changes on a stable branch


FelixKHWong:
Will fix formatting once approval is given, otherwise formatting changes will clutter the PR

FelixKHWong:
Deprecated for https://github.com/nightingale-project/nightingale/pull/57 which integrates directly into main

PR #57, opened 2023-02-27
User/felix/pose estimation w errorhandle

FelixKHWong:
Final pose estimation PR with integration into mission planner and error handling (retries on pose retrieval/estimation and fallback point if estimation is not made)
See https://github.com/nightingale-project/nightingale/pull/49 for more details on pose estimation node


nightingale_perception/src/nightingale_perception/pose_estimation.py
```diff
@@ -0,0 +1,391 @@
...
+        new_point_stamped = PointStamped()
+        new_point_stamped.header.frame_id = point_stamped.header.frame_id
+        new_point_stamped.point.x = point_vals[0] * self.MM_TO_M
+        new_point_stamped.point.y = point_vals[1] * self.MM_TO_M
+        new_point_stamped.point.z = point_vals[2] * self.MM_TO_M
+
+        return new_point_stamped
+
+    def transform_point_baselink(self, point_stamped):
+        while not rospy.is_shutdown():
```
younesr1, line 368: nit: consider a sleep or a rate to not loop super fast

nightingale_perception/src/nightingale_perception/pose_estimation.py
```diff
@@ -0,0 +1,391 @@
+#!/usr/bin/env python
```
younesr1, line 1: impressive stuff!

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -38,6 +41,11 @@ def __init__(self):
 
         self.phases = queue.Queue()
 
+        self.fallback_bin_goal = Point()
```
younesr1: hmm this adds yet another hard-coded point. we are starting to have quite a lot of thse in several places, and theyre hard to keep track of and update. do u think itd be reasonable to delete this fallback_bin_goal Point and have the robot call extend_restock() if the dynamic detection fails? extend restock actually has this point defined inside of iirc

younesr1 (dismissed):
great work on this felix!

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: sure sounds good

nightingale_perception/src/nightingale_perception/pose_estimation.py
```diff
@@ -0,0 +1,392 @@
...
+                self.result.pose_est_fail = 2
+                self.pose_estimation_server.set_aborted(self.result)
+                return
+            else:
+                rospy.loginfo(f"Need valid target")
+                self.result.pose_est_fail = 3
+                self.pose_estimation_server.set_aborted(self.result)
+
+    def preempt_cb(self):
+        pass
```
bdschnapp, line 183: raise NotImplemented

bdschnapp (commented):
I did not do an in-depth review as I am not familiar with much of what is going on, but over all looks good and I believe the integration with mission planner and manipulation is correct

nightingale_perception/src/nightingale_perception/pose_estimation.py
FelixKHWong: I realized this node returns so fast that it's pretty much a service so I don't think this will ever be called

nightingale-project (approved):

PR #61, opened 2023-03-01
add workspace check for handoff

younesr1:


nightingale_dispatcher/src/nightingale_dispatcher/move_arm_task.py
```diff
@@ -4,13 +4,43 @@
...
+import tf2_ros
+import numpy as np
 
 
 class MoveArmTask(Task):
     def __init__(self):
         self.manipulation = ManipulationControl()
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+        self.right_arm_length = 1.5  # younes todo check measure
```
FelixKHWong: possibly use the distance to the wrist since the important distance is from shoulder to the end effector(the padded part inside the hand). Using this distance will give a conservative estimate to get a good response. Also if we are locking the wrist orientation to be sideways then the distance is probably going to have to be even smaller.

nightingale_dispatcher/src/nightingale_dispatcher/move_arm_task.py
```diff
@@ -4,13 +4,43 @@
...
 
     def execute(self, end_pose):
         assert len(end_pose) is 7
         self.manipulation.jnt_ctrl.cmd_right_arm(end_pose)
         return TaskCodes.SUCCESS
+
+    # this function returns true if a given point is within the right arm's workspace
+    def witihin_workspace(self, point):
```
FelixKHWong, line 38: I feel like workspace isn't as descriptive as using like "within_reach"

nightingale-project: i mean workspace in the robotics sense which has a specific definition: https://modernrobotics.northwestern.edu/nu-gm-book-resource/2-5-task-space-and-workspace/#:~:text=The%20workspace%20is%20a%20specification,has%20the%20workspace%20shown%20here.

bdschnapp (dismissed):

FelixKHWong (dismissed):

nightingale-project (approved):

PR #62, opened 2023-03-02
User/le2tang/torso head joint ctrl

le2tang:
Add new class BodyControl with home() and handoff() functions that move torso and head into the home and handoff positions defined in the robot configuration yaml


nightingale_body/src/nightingale_body/body_control.py
```diff
@@ -0,0 +1,115 @@
...
+        rospy.loginfo(f"Manipulation Control: waiting for robot configuration service")
+        rospy.wait_for_service(robot_config_service)
+        lookup_client = rospy.ServiceProxy(
+            robot_config_service, RobotConfigurationLookup
+        )
+        rospy.loginfo(
+            "Nightingale Manipulation Control found robot configuration lookup service server"
+        )
+        try:
+            response = lookup_client("home", RobotConfigurationLookupRequest.HEAD)
```
FelixKHWong, line 73: Is it necessary to run these functions on init? I'm just scared something goes wrong and breaks the tablet holder. Also is this going to make the torso go high on init?

nightingale_utils/configs/robot_configuration_lookup.yaml
```diff
@@ -29,7 +37,7 @@ joint_configurations: {
             joints: [0.23, 0.23, 0.23],
         },
         right_gripper: {
-            names: ["right_gripper_finger1_joint", "right_gripper_finger2_joint", "right_gripper_finger3_joint" ]
+            names: ["right_gripper_finger1_joint", "right_gripper_finger2_joint", "right_gripper_finger3_joint" ],
```
FelixKHWong, line 40: not sure if these should be here but doesn't seem to break anything so we'll keep

FelixKHWong (changes requested):

nightingale_body/src/nightingale_body/body_control.py
le2tang: No movement occurs on init, just looking up the values for the desired states

nightingale_utils/configs/robot_configuration_lookup.yaml
le2tang: They are necessary to run the tests, so it's worth keeping here

nightingale-project (approved):
nice!

FelixKHWong (approved):

PR #63, opened 2023-03-02
Fix service call

le2tang:
There were some mistakes in the service call for the manipulator robot configurations


FelixKHWong (approved):
good catch!

nightingale-project (approved):

PR #65, opened 2023-03-03
User/le2tang/dispatcher demo

le2tang:
Add field to control which phase is started in the mission plan.
Tested using dummy tasks, needs full run through test

### DIFF
--- a/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
+++ b/nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
@@ -54,7 +54,7 @@ class MissionPlanner:
         # TODO: first go to doorside then bedside when door opening added
 
         # update to driving screen
-        task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
+        task_response = self.send_interface_request_task.execute(RobotStatus.DRIVING)
 
         status = self.navigate_task.execute(self.room, "bedside")
         if status == TaskCodes.ERROR:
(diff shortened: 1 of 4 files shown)


nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -31,20 +31,30 @@ def __init__(self):
...
-        self.phases = queue.Queue()
-
     def go_to_patient_phase(self):
         rospy.loginfo("Nightingale Mission Planner going to patient")
         # Assume door is open
         # TODO: first go to doorside then bedside when door opening added
 
         # update to driving screen
-        task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
+        task_response = self.send_interface_request_task.execute(RobotStatus.DRIVING)
```
FelixKHWong, line 60: why is this a change lol

FelixKHWong (dismissed):

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
le2tang: I needed to run a full test, and there was another instance where this was spelled wrong that could prevent a full test, so i did a find and replace.

FelixKHWong: oh I did not see that nice

bdschnapp (commented):
I would like to see this tested IRL before approving, but otherwise it looks good to me

PR #66, opened 2023-03-03
User/le2tang/body control itegration

le2tang:
Add task in mission planner to raise and lower torso as needed

### DIFF
--- a/nightingale_body/src/nightingale_body/body_control.py
+++ b/nightingale_body/src/nightingale_body/body_control.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import actionlib
+import numpy as np
 import rospy
 
 from moveit_action_handlers.msg import (
@@ -8,14 +9,17 @@ from moveit_action_handlers.msg import (
     MoveToJointsMoveItGoal,
     PropertyValuePair,
 )
-
 from nightingale_msgs.srv import (
     RobotConfigurationLookup,
     RobotConfigurationLookupRequest,
 )
+from sensor_msgs.msg import JointState
 
 
 class BodyJointControl:
+    CLOSE_DIST_TOL = 1e-2  # m
+    CLOSE_ANGLE_TOL = 0.087  # rad = 5 deg
+
     def __init__(self):
         self.head_ac = actionlib.simple_action_client.SimpleActionClient(
             "/moveit_action_handlers/head/joint_ctrl", MoveToJointsMoveItAction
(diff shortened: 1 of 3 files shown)


nightingale_body/src/nightingale_body/body_control.py
```diff
@@ -46,17 +50,55 @@ def _make_goal(self, jnt_name, jnt_pos, ee_vel=0.5, ee_accel=0.5, timeout=5):
...
+    def head_is_close(self, target_jnt_name, target_jnt_pos):
+        curr_jnt_state = rospy.wait_for_message("/movo/head/joint_states", JointState)
+        curr_jnt_pos = {
+            name: pos for name, pos in zip(curr_jnt_state.name, curr_jnt_state.pos)
+        }
+
+        diff = np.zeros(len(target_jnt_name))
+        for idx, (tgt_name, tgt_pos) in enumerate(zip(target_jnt_name, target_jnt_pos)):
+            try:
+                diff[idx] = np.abs(curr_jnt_pos[tgt_name] - tgt_pos)
```
FelixKHWong, line 79: so you are taking the angle error of each joint and seeing if the sum is too large? Wouldn't we want to check each angle against the error tolerance rather than the sum? I'm thinking that one joint could have low error but the other has high error but still within bound. Would it be better to have a tighter tolerance and check each joint for lower error?

nightingale_body/src/nightingale_body/body_control.py
```diff
@@ -46,17 +50,55 @@ def _make_goal(self, jnt_name, jnt_pos, ee_vel=0.5, ee_accel=0.5, timeout=5):
...
+    def torso_is_close(self, target_jnt_name, target_jnt_pos):
+        curr_jnt_state = rospy.wait_for_message(
+            "/movo/linear_actuator/joint_states", JointState
+        )
+        curr_jnt_pos = {
+            name: pos for name, pos in zip(curr_jnt_state.name, curr_jnt_state.pos)
+        }
+
+        diff = np.zeros(len(target_jnt_name))
+        for idx, (tgt_name, tgt_pos) in enumerate(zip(target_jnt_name, target_jnt_pos)):
```
FelixKHWong, line 94: Isn't there only a single joint here? Is this loop necessary for the torso?

bdschnapp: If we are being consistent, the joint target passed in will be a list (with 1 value) in which case the loop is the easiest solution even though it only gets looped through once

nightingale_body/src/nightingale_body/body_control.py
bdschnapp: I agree with this. the movement should be executed if any joints are outside of a tolerance, not if the sum of error is outside of a tolerance.

nightingale_body/src/nightingale_body/body_control.py
```diff
@@ -46,17 +50,55 @@ def _make_goal(self, jnt_name, jnt_pos, ee_vel=0.5, ee_accel=0.5, timeout=5):
         return goal
 
     def cmd_head(self, target_jnt_name, target_jnt_pos):
```
bdschnapp, line 52: cmd_head (as well as cmd_torso) should take only the target joint values. This is for consistency with arm joint control.

nightingale_body/src/nightingale_body/body_control.py
```diff
@@ -1,21 +1,25 @@
 #!/usr/bin/env python3
 
 import actionlib
+import numpy as np
 import rospy
```
bdschnapp, line 6: import nightingale_manipulation.manipulation_action_client.joint_goal instead of redefining it with the _make_goal function. I could also move joint_goal into a nightingale_manipulation.utils file if that would look better

bdschnapp (changes requested):
Over all well done, just a few small changes requested

nightingale_body/src/nightingale_body/body_control.py
bdschnapp: didnt mean to specifically highlight this line here sorry, it wouldn't let me highlight the _make_goal function definition for some reason.

le2tang: I agree that the goal definition function shouldn't really be repeated here, but I'm not sure adding this dependency is wise since broke out the body control from the manipulation to remove any dependencies between the two.

nightingale_body/src/nightingale_body/body_control.py
le2tang: I did this to separate the responsibility of getting and storing the desired values from calling the action client. The BodyControl class gets the joint names along with the desired values, so it seems to make sense to pass them from the BodyControl class rather than storing the joint names in the joint control class

nightingale-project (dismissed):

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -46,6 +48,9 @@ def go_to_patient_phase(self):
         # update to driving screen
         task_reponse = self.send_interface_request_task.execute(RobotStatus.DRIVING)
 
+        rospy.loginfo("Lowering body to home")
+        status = self.move_body_task.home()
```
nightingale-project, line 65: notice that the return is being stored and immediately overwrriten. so this line is the same as just `self.move_body_task.home()`.

not a big deal, but just in case you thought u were using the return code

nightingale-project (dismissed):

bdschnapp (approved):

nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
FelixKHWong: @le2tang I'm pushing this through but take note of this

PR #69, opened 2023-03-04
march 4 work

nightingale-project:


nightingale_dispatcher/src/nightingale_dispatcher/mission_planner.py
```diff
@@ -198,25 +221,26 @@ def handoff_items_phase(self):
...
-        #    status = self.move_arm_task.extend_restock()
-        # else:
-        #     status = self.move_arm_task.extend_handoff(bin_goal_pt)
-        # if status != TaskCodes.SUCCESS:
-        if self.move_arm_task.extend_restock() != TaskCodes.SUCCESS:
+        if status != TaskCodes.SUCCESS or not self.move_arm_task.within_workspace(
+            bin_goal_pt
+        ):
+            rospy.logwarn("UNABLE TO FIND POSE. FALLING BACK TO SAFE HANDOFF POSITION")
+            status = self.move_arm_task.extend_restock()
```
FelixKHWong, line 241: We should probably redefine this hardcoded extend even though we should never get to it. Maybe instead create a new default extend which just lifts the box up?

le2tang (dismissed):

bdschnapp (dismissed):

PR #81, opened 2023-03-14
Create new package for symposium demo

le2tang:
Add a package to hold all the code specifically for the symposium demo

### DIFF
--- a/nightingale_demo/CMakeLists.txt
+++ b/nightingale_demo/CMakeLists.txt
@@ -1,0 +1,206 @@
+cmake_minimum_required(VERSION 3.0.2)
+project(nightingale_demo)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+# add_compile_options(-std=c++11)
+
+## Find catkin macros and libraries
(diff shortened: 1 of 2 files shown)


FelixKHWong:
Do we need a new demo package? Could we just make a new branch and modify the current mission planner? I was thinking we would use the original mission planner but modify it so we only use the handoff state and change what is inside it. This way we can follow the same start up procedure that we are all familiar with without making many new changes.

younesr1:
> Do we need a new demo package? Could we just make a new branch and modify the current mission planner? I was thinking we would use the original mission planner but modify it so we only use the handoff state and change what is inside it. This way we can follow the same start up procedure that we are all familiar with without making many new changes. 

I think we should use a new package thats simpler and more boiler plate. We dont wanna inherit all the complexity if MP. We just want a simple main.py in a new pkg

nightingale-project (approved):
nit: consider calling this nightingale_symposium since the demo was for the profs and we have branches calling that the demo

PR #83, opened 2023-03-16
User/le2tang/proprioception

le2tang:
Add collision detection under the package nightingale_proprioception
Add calibration script to move to random positions and get the torque calibration constants and the collision model covariance matrix


nightingale_proprioception/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
...
+project(nightingale_proprioception)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+# add_compile_options(-std=c++11)
+
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  geometery_msgs
```
FelixKHWong: I think this should be geometry. Did this build successful with this? Or is this not a typo?

nightingale_proprioception/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
...
+)
+
+## System dependencies are found with CMake's conventions
+# find_package(Boost REQUIRED COMPONENTS system)
+
+
+## Uncomment this if the package has a setup.py. This macro ensures
+## modules and global scripts declared therein get installed
+## See http://ros.org/doc/api/catkin/html/user_guide/setup_dot_py.html
+# catkin_python_setup()
```
FelixKHWong, line 24: Might need to uncomment this line

nightingale_proprioception/src/collision_detection.py
```diff
@@ -0,0 +1,306 @@
...
+
+        self.iter = 0
+        self.iter_decimation = 10
+
+        self.joint_state_sub = rospy.Subscriber(
+            f"/movo/{self.arm_side}_arm/joint_states", JointState, self.joint_state_cb
+        )
+
+    def joint_state_cb(self, msg):
+        if self.iter == self.iter_decimation - 1:
```
FelixKHWong, line 237: What's the iter_decimation parameter for?

nightingale_proprioception/CMakeLists.txt
FelixKHWong: also in other lines if this is indeed a typo

FelixKHWong (commented):
Good work! I'm not much help debugging and giving feedback on the math but I think it's impressive! How long would it take to calibrate the right arm for collision detection?

nightingale_proprioception/src/collision_detection.py
le2tang: This parameter lowers the running rate of the algorithm. The algorithm takes more than 10 ms to run, so this parameter slows the input rate down by only running once every `iter_decimation` times a joint state is published.

nightingale_proprioception/CMakeLists.txt
le2tang: I don't have a setup.py in my package since each script is independent of the other, so uncommenting this line leads to a build error

FelixKHWong (approved):

PR #84, opened 2023-03-16
User/felix/symposium tablet

FelixKHWong:
If we are not making a new branch for the symposium I will need to clean things up a bit to make things work better.

### DIFF
--- a/nightingale_interface/main.py
+++ b/nightingale_interface/main.py
@@ -59,15 +59,17 @@ class MainApp(MDApp, ScreenWrapper):
 
     def main(self):
         self._other_task = asyncio.ensure_future(self.backend())
-        self._wd_task = asyncio.ensure_future(self.watchdog_run())
+        #self._wd_task = asyncio.ensure_future(self.watchdog_run())
 
         async def run_wrapper():
             await self.async_run(async_lib="asyncio")
             print("App done")
             self._other_task.cancel()
-            self._wd_task.cancel()
+            #self._wd_task.cancel()
 
-        return asyncio.gather(run_wrapper(), self._other_task, self._wd_task)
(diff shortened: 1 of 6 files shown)


nightingale_interface/main.py
```diff
@@ -39,6 +42,8 @@ class MainApp(MDApp, ScreenWrapper):
     ros_action_topic = None
     interface_screen_topic = None
     estop_topic = None
+    arm_topic = None
```
FelixKHWong: delete these not necessary

nightingale_interface/main.py
```diff
@@ -57,17 +62,20 @@ class MainApp(MDApp, ScreenWrapper):
     pending_action = "0"
     payload = ""
 
+    arm_collision = False
```
FelixKHWong: delete

nightingale_ros_bridge/src/nightingale_ros_bridge/bridge_interface_config.py
```diff
@@ -10,13 +10,16 @@ class RobotStatus:
...
     ARM_RETRACTED = 8
+    ARM_COLLISION = 9
 
 
 class BridgeConfig:
     USER_INPUT_TOPIC = "nightingale_ui/user_input"
     ROBOT_STATUS_TOPIC = "nightingale_ui/robot_status"
     ESTOP_TOPIC = "nightingale_ui/estop_status"
     UPDATE_UI_SERVICE = "nightingale_ui/update_ui_state"
+    EXTEND_ARM_TOPIC = "nightingale_ui/arm_status"
```
FelixKHWong: delete these since we can just use the user input and update ui state topics

